PR #2, opened 2019-09-25
Added main view

auderephilip:
This creates a set of files for us to work off of.

* corestore.ts will centralize all access to supposedly-server data.  We can jam fake data in here for now to unblock UI development.
* MainChrome.tsx presents the main site chrome.  Haven't integrated it into LoginScreen because we might want that one screen to look different.
* Added react-tabs to support basic tabs.  The idea here is that one person could play many roles, in which case we'd want the app to haev an easy way to flip between that person's tasks, sorted by role (one role per tab).

Everything looks nasty right now, but it's because I've spent zero time on styling.  Perhaps a more important question is what we want to do about CSS in React;  in RN, CSS can be easily embedded into each component's file using StyleSheet.create().  But I don't see an equivalent in React.  For now, I've put individual CSS files using the same name as the class that consumes them;  but the problem is that these styles are technically app-wide (meaning you can easily get into name collisions across components, which smells bad).  Here are a few proposals:

* Do we want a /styles/ directory?  For now, keeping the CSS in the same place as the component feels better... but I could see that feeling messy over time.
* Do we potentially want just one big CSS, so that name collisions are obvious and avoidable (if we alphabetize)?
* Or perhaps do we want a naming convention, like "every component's styles should all fall under a parent style for that component," like `.myComponent .myStyle`

LMK thoughts.

image


auderephilip:
@jennya / @balderdash : wondering if either of you might have opinions about the right way to handle CSS in a React project.

auderephilip:
Here's a scoop from the interwebs about how to solve this CSS problem:  essentially, create individual CSS files and import them, as I've already done, but also enforce naming conventions that always have the component name as prefixes.  So, for instance, MyComponent.tsx would import MyComponent.css, and only use classNames that begin with `MyComponent_` (e.g. `MyComponent_container`).

This basically adds pseudo-namespace-like behavior to big-app CSS that avoids most collisions.

I'll make that change and ping it back to this PR.

auderephilip:
Ok, now I've implemented the preview pane on the left using test data.  This should hopefully get us to a point where we can break off a bunch of work in parallel... like implementing the details view, pulling in the photos from URLs, implementing commits of Notes changes, etc.

image

And here's the Ops view:
image

src/store/corestore.ts
```diff
@@ -7,9 +8,153 @@ const FIREBASE_CONFIG = {
...
+  totalCost: number;
+  claimedCost: number;
+  site: Site;
+  timestamp: number;
+  notes?: string;
+};
+
+export type ReimbursementTask = {
+  site: Site;
+  claims: Task[];
```
audererob: Nit: could this be ClaimTask[] ?

auderephilip: Great question, @audererob .  I had started there, but wondered whether the Payor would want to know when each claim was actually approved.  Not positive that's needed, and in fact, it might be over-engineering.  LMK.

audererob: Thanks, that makes sense - I deleted my question when I realized the extra data could prove useful (as an FYI for people in the future who are sure to look at this, my question was whether Task[] should be ClaimTask[]).

src/Components/TaskList.tsx
```diff
@@ -0,0 +1,53 @@
...
+      return;
+    }
+    this.setState({ selectedIndex: index });
+  };
+
+  render() {
+    return (
+      <div className={this.props.className}>
+        <div className="tasklist_header">Items to Review</div>
+        <div className="tasklist_items">
```
audererob: Nit: there isn't a style declared for tasklist_items.

auderephilip:
Thanks, eagle-eye @audererob !  Removed unused style, and merged.  Hopefully this'll let everyone start working in parallel!

goldjay:
@auderephilip For CSS in React a lot of people use [Emotion](https://emotion.sh/docs/introduction) or [Styled Components](https://www.styled-components.com/). Both of which link styles to specific components sort of how it's done in React-Native.

auderephilip:
Ooh, interesting.  Thanks for sending those, @goldjay ;  TIL.

Both look a bit nasty to me, TBH... but as long as either/both support the idea of constants (e.g. we'd want to be able to share, say, DEFAULT_PADDING between components), we should be ok.

I'm happy to go with whatever decision the team wants on this.  Just LMK what everyone decides, and I'll make sure future PRs conform.

auderephilip:
One thing that's not clear to me, for instance, is how to compose styles together (such that a component can not only have its own style, but take a `style` prop that it merges into its toplevel container in order to let consumers override the default styling).  There's some className concatting that is doable... but I'm not sure how that plays with other inline CSS frameworks we may want to adopt.

But I know literally zero about this space.  I'll gladly go with whatever the team recommends.

PR #3, opened 2019-09-27
Adds CSV upload capability

auderephilip:
The UI looks nasty, and has zero investment in it so far because I anticipate we'll change it.  Just introduces an icon which, when clicked, uses an Input object to upload a CSV to Firebase Storage.

This is a temporary solution, as the backend of the project is coming together.  Uploading to Firebase Storage allows us to next use Storage Triggers (as Cloud Functions) to process each upload into a DB (which is also temporary, until our backend is up and running).

image

And an uploaded file:
image


src/Components/TopBar.tsx
```diff
@@ -1,17 +1,76 @@
...
+    } catch (e) {
+      alert(e);
+    }
+  };
 
-class TopBar extends React.PureComponent {
   render() {
+    const { roles, showFileSelector } = this.state;
+    const uploadButton =
+      roles.includes(UserRole.AUDITOR) && !showFileSelector ? (
```
audererob: Something for us to consider going forward is whether we'd want Roles to be comprised of "allowed capabilities" to avoid directly mapping role type to UX. That would avoid us having to update a bunch of code in case a new role type came along (e.g. senior auditor vs. junior auditor). This code would instead be something more like const uploadButton = roles.hasUserCapability(UserCapabilities.CSVUPLOAD)...

auderephilip: Great point, and agreed 100%.  Happy to mod/dork-around-with as we discuss more.

It'll be interesting to chat, longer-term, about whether the UI itself pivots by role vs. capability...

PR #4, opened 2019-09-28
Introduces server-side CSV processing

auderephilip:
While the real backend is being written, this uses a Firebase Cloud Function storage trigger to parse uploaded CSVs into two areas:

* A store of all rows uploaded, lumped into per-CSV buckets called "batches"
* A set of sampled rows which then become workflow items for Auditors

All of this storage and code is temporary, but should continue to unblock frontend development.  Since this is a storage trigger, you can run / test the code just by using Firebase Console to upload a CSV into the csvuploads directory in Firebase Storage.

You'll need flows-app-staging-key.json on your local machine in order to read from Firebase Storage via this cloud function.  It's stored in LastPass.

Next steps are to change `corestore.ts` in the React app to pull real workitem data from Firestore (and eventually from AWS).

Here is what the actual parsed CSV looks like in Firestore:
image

And then the auditor tasks:
image

PR from 32,000 ft in the air!  Gotta love in-flight wifi.  It's magic.


auderephilip:
@balderdash : this is the PR that's not yet merged, which we discussed changing today (such that each ClaimTask represents one pharmacy).

The change to this PR would be around what type of object is written into auditor_todo... we'll likely need to change corestore.ts as well to reflect the new aggregated-per-pharmacy ClaimTask .

auderephilip:
(merged after consultation with @balderdash to unblock both of us.  Edits and updates welcome in subsequent PRs).

PR #5, opened 2019-09-28
Pull auditor tasks from server

auderephilip:
We can now pull auditor tasks directly from the server, since CSV upload is working.  As we develop Details View a bit more, we'll need to pull more and more data into the actual ClaimTask;  but for now, this replaces our test data with data that came from a CSV.

I haven't updated the Payor view with this data because we'd likely want to enable a real edit flow instead of doing that (i.e. we'll likely want to implement the "click Approve" functionality in order to populate the Payor view with real data).

Here's the data they gave us, pulled from Firestore:
image

### DIFF
--- a/src/store/corestore.ts
+++ b/src/store/corestore.ts
@@ -39,6 +39,7 @@ export type ClaimTask = {
   patientFirstName: string;
   patientLastName: string;
   patientID?: string;
+  phone?: string;
   item: string;
   totalCost: number;
   claimedCost: number;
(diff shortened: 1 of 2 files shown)


src/store/corestore.ts
```diff
@@ -91,6 +100,37 @@ export async function tasksForRole(role: UserRole): Promise<Task[]> {
...
+      photoIDUri: d["g4:B03.1 Photo of ID card"],
+      photoMedUri: d["g5:B04 (Medication)"],
+      photoMedBatchUri: d["g5:B05 (Medication batch)"],
+      item: d["Type received"],
+      totalCost: d["Total med price covered by SPIDER"],
+      claimedCost: d["Total reimbursement"],
+      timestamp: new Date(d["YYYY"], d["MM"], d["DD"]).getTime(),
+      site: {
+        name: d["g3:B01 Pharmacy name"],
+        phone: "+254 867 5309"
```
audererob: Is the 'phone' field not yet available?

auderephilip: Oooh.  Good point on the phone bit.  Just added.  thanks!

PR #6, opened 2019-10-01
Details Views

goldjay:
* Basic UI for details pages
* Image Zoom

Next Steps:
* Re-do details UI on a by-pharmacy basis once the data shape is changed
* Search claims

Details Views Demo

Image Zoom

### DIFF
--- a/src/Components/DataTable.tsx
+++ b/src/Components/DataTable.tsx
@@ -1,0 +1,39 @@
+import React from "react";
+import "./DataTable.css";
+
+interface Props {
+  data: any[];
+}
+
+const DataTable = (props: Props) => {
+  const columns = Object.keys(props.data[0]);
+
+  return (
+    <table className="datatable_table">
+      <thead className="datatable_th">
+        <tr>
+          {columns.map((item, idx) => {
+            return <th key={item + idx}>{item}</th>;
+          })}
+        </tr>
+      </thead>
+      <tbody>
+        {props.data.map((item, index) => {
+          return (
+            <tr key={item + index}>
+              {columns.map((col, idx) => {
+                return (
+                  <td className="datatable_td" key={col + idx}>
+                    {item[col]}
+                  </td>
+                );
+              })}
+            </tr>
+          );
+        })}
+      </tbody>
+    </table>
+  );
+};
+
+export default DataTable;
--- a/src/Components/DataTable.tsx
+++ b/src/Components/DataTable.tsx
@@ -6,6 +6,9 @@ interface Props {
 }
 
 const DataTable = (props: Props) => {
+  if (props.data.length === 0) {
+    return null;
+  }
   const columns = Object.keys(props.data[0]);
 
   return (
--- a/src/Components/TaskList.tsx
+++ b/src/Components/TaskList.tsx
@@ -1,11 +1,13 @@
 import React from "react";
 import { Task } from "../store/corestore";
+import LabelWrapper from "./LabelWrapper";
 import "./TaskList.css";
 
 type Props = {
   tasks: Task[];
   renderItem: (task: Task, isSelected: boolean) => JSX.Element;
   className?: string;
+  onSelect?: (index: number) => void;
 };
 
 type State = {
@@ -26,13 +28,16 @@ class TaskList extends React.Component<Props, State> {
     if (isNaN(index)) {
       return;
     }
+    if (!!this.props.onSelect) {
+      this.props.onSelect(index);
+    }
+
     this.setState({ selectedIndex: index });
   };
 
   render() {
     return (
-      <div className={this.props.className}>
-        <div className="tasklist_header">Items to Review</div>
+      <LabelWrapper className={this.props.className} label={"Items to Review"}>
         <div>
           {this.props.tasks.map((task, index) => {
             return (
@@ -45,7 +50,7 @@ class TaskList extends React.Component<Props, State> {
             );
           })}
         </div>
-      </div>
+      </LabelWrapper>
     );
   }
 }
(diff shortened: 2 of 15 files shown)


src/Components/TaskList.tsx
```diff
@@ -26,13 +28,16 @@ class TaskList extends React.Component<Props, State> {
...
+    }
+
     this.setState({ selectedIndex: index });
   };
 
   render() {
     return (
-      <div className={this.props.className}>
-        <div className="tasklist_header">Items to Review</div>
+      <LabelWrapper className={this.props.className} label={"Items to Review"}>
```
auderephilip: Not sure I really understand the existence of LabelWrapper.  What are the benefits of doing this?

src/Screens/MainView.tsx
```diff
@@ -48,6 +58,86 @@ class MainView extends React.Component<Props, State> {
...
+    this.setState({ notes: "" });
+  }
+
+  _renderClaimDetails = (task: Task) => {
+    const claim = task as ClaimTask;
+
+    const claimImageURLs = [
+      "https://vetstreet-brightspot.s3.amazonaws.com/7d/8b/98ea05c4403e8d4163dc8c8991c0/kitten-in-bed-thinkstockphotos-466265898.jpg",
+      "https://vetstreet-brightspot.s3.amazonaws.com/7d/8b/98ea05c4403e8d4163dc8c8991c0/kitten-in-bed-thinkstockphotos-466265898.jpg",
+      "https://vetstreet-brightspot.s3.amazonaws.com/7d/8b/98ea05c4403e8d4163dc8c8991c0/kitten-in-bed-thinkstockphotos-466265898.jpg"
```
auderephilip: Real images now available, I believe, from PR #5 getting merged.

src/Components/TextItem.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface Data {
+  [key: string]: string;
+}
+
+interface Props {
+  data: Data;
+}
+
+const TextItem = (props: Props) => {
```
auderephilip: Not sure I understand this either, given that it's such a simple thing.  Is the idea to have a single semantic object represent a key:value pair when shown to the user?

src/Screens/MainView.tsx
```diff
@@ -48,6 +58,86 @@ class MainView extends React.Component<Props, State> {
...
+          data={{
+            Patient: `${claim.patientFirstName} ${claim.patientLastName}`
+          }}
+        />
+        <TextItem data={{ Item: claim.item }} />
+        <ImageRow imageURLs={claimImageURLs} />
+        <LabelTextInput onTextChange={this._onNotesChanged} label={"Notes"} />
+        <div className="mainview_button_row">
+          <Button label="Decline" onClick={() => console.log("CLICKED")} />
+          <Button label="Approve" onClick={() => console.log("CLICKED")} />
```
auderephilip: Could these just go to `_onApprove` and `_onDecline`?

src/Screens/MainView.tsx
```diff
@@ -98,7 +206,20 @@ class MainView extends React.Component<Props, State> {
...
+          return (
+            <TabPanel
+              key={"tab_" + index}
+              style={{
+                display: "flex",
+                flexDirection: "row"
+              }}
+            >
+              {this._renderRolePanel(index)}
+              {this._renderDetailsPanel(index)}
```
auderephilip: We might want to rename these something other than "Panel", since <TabPanel> has a specific meaning here (i.e. a panel should be something that renders the entire area of a tab's client region).  Perhaps `_renderRolePane` and `_renderDetailsPane`?

src/Components/TaskList.tsx
goldjay: Easier consistent styling in both places. I assume they would be matching still down the road and there's a possibility of adding more panes.

src/Components/TextItem.tsx
goldjay: Yes, that's the idea. Ideally we'd be able to shape the data into an object and just plug that into a larger component, which would generate a column of TextItems.

src/Screens/MainView.tsx
goldjay: _renderRolePane and _renderDetailsPane sound good.

src/Components/DataTable.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import "./DataTable.css";
+
+interface Props {
+  data: any[];
+}
+
+const DataTable = (props: Props) => {
+  const columns = Object.keys(props.data[0]);
```
audererob: Nit: props.data might not have any elements in it, which would cause this to crash.

src/Screens/MainView.tsx
```diff
@@ -48,6 +58,94 @@ class MainView extends React.Component<Props, State> {
...
+        <TextItem
+          data={{
+            Patient: `${claim.patientFirstName} ${claim.patientLastName}`
+          }}
+        />
+        <TextItem data={{ Item: claim.item }} />
+        <ImageRow imageURLs={this._extractImageURLs(claim)} />
+        <LabelTextInput onTextChange={this._onNotesChanged} label={"Notes"} />
+        <div className="mainview_button_row">
+          <Button label="Decline" onClick={() => this._onDecline()} />
```
audererob: I think it prevents DOM re-renders if you declare _onDecline and _onApprove with capture rather than here, because they otherwise become anonymous functions which will have a different value each time through. This would instead become onClick={this._onDecline}, a constant value.

auderephilip: Yah - important in render not to have fat-arrow functions as handlers.

src/Components/DataTable.tsx
goldjay: Good point. I'll add a check.

PR #7, opened 2019-10-01
More ClaimTask Properties

goldjay:
* Added age and sex to ClaimTask (They were in the Sketch spec, but not on ClaimTask)
* Fixed patientID key

Screen Shot 2019-10-01 at 3 15 22 PM

### DIFF
--- a/src/Screens/MainView.tsx
+++ b/src/Screens/MainView.tsx
@@ -84,6 +84,8 @@ class MainView extends React.Component<Props, State> {
 
   _renderClaimDetails = (task: Task) => {
     const claim = task as ClaimTask;
+    const age = !!claim.patientAge ? claim.patientAge : "";
+    const sex = !!claim.patientSex ? claim.patientSex : "";
 
     return (
       <LabelWrapper label="DETAILS VIEW">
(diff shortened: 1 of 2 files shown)


src/Screens/MainView.tsx
```diff
@@ -84,6 +84,8 @@ class MainView extends React.Component<Props, State> {
 
   _renderClaimDetails = (task: Task) => {
     const claim = task as ClaimTask;
+    const age = !!claim.patientAge ? claim.patientAge : "";
+    const sex = !!claim.patientSex ? claim.patientSex : "";
```
audererob: In the event one or both are missing, perhaps do something like:

let patientProps = [];
if (!!claim.patientAge && claim.patientAge!.length > 0) patientProps.push(claim.patientAge);
if (!!claim.patientSex && claim.patientSex!.length > 0) patientProps.push(claim.patientSex);
const patientInfo = patientProps.length > 0 ? ` (${patientProps.join(", ")})` : "";

...then append patientInfo to the Patient: string value in the TextItem component.

goldjay: Nice! I'll make the change.

PR #8, opened 2019-10-01
Group claims by pharmacy

balderdash:
This PR modifies the function that parses CSVs to just create one task per pharmacy, with a sampled set of entries for that pharmacy included in the task, and updates the task list and details view for `ClaimTask` to render appropriately.

The task list entries are super barebones, we'll probably want to add more detail here (and presumably in real CSVs the "pharmacy name" column won't just be a number):
image

The details might need to be collapsible or something:
image

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -86,19 +86,17 @@ async function completeCSVProcessing(cache: any[]) {
 
   // Now generate Auditor work items representing each sampled row.  No need to
   // await this stuff either -- we just let each one happen whenever they do.
-  sampledRowsByPharmacy.forEach(pharm => {
+  sampledRowsByPharmacy.forEach(async pharm => {
     console.log(`Pharmacy ${pharm.key} has ${pharm.values.length} rows`);
-    pharm.values.forEach(row =>
-      admin
-        .firestore()
-        .collection(AUDITOR_TODO_COLLECTION)
-        .doc()
-        .set({
-          batchID,
-          pharmacyID: pharm.key,
-          data: row
-        })
-    );
+    await admin
+      .firestore()
+      .collection(AUDITOR_TODO_COLLECTION)
+      .doc()
+      .set({
+        batchID,
+        pharmacyID: pharm.key,
+        data: pharm.values
+      })
   });
   console.log(
     `Seem to have processed ${sampledRowsByPharmacy.length} pharmacies`
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -84,9 +84,8 @@ async function completeCSVProcessing(cache: any[]) {
     };
   });
 
-  // Now generate Auditor work items representing each sampled row.  No need to
-  // await this stuff either -- we just let each one happen whenever they do.
-  sampledRowsByPharmacy.forEach(async pharm => {
+  // Now generate Auditor work items representing each sampled row.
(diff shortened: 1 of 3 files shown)


src/Screens/MainView.tsx
```diff
@@ -115,31 +123,29 @@ class MainView extends React.Component<Props, State> {
 
   _renderReimbursementDetails = (task: Task) => {
     const reimbursement = task as ReimbursementTask;
-    if (!reimbursement || !reimbursement.claims) return null;
-    const claimAmounts = reimbursement.claims.map(task => {
-      const claim = task as ClaimTask;
-      return claim.claimedCost;
+    const claim = reimbursement.claim as ClaimTask;
+    if (!reimbursement || !claim || !claim.entries) return null;
```
auderephilip: nit: prefer even one-line blocks to have braces:
```
if (!reim || !claim || !claim.boo) {
  return null;
}
```

functions/src/index.ts
```diff
@@ -86,19 +86,17 @@ async function completeCSVProcessing(cache: any[]) {
...
-    );
+    await admin
+      .firestore()
+      .collection(AUDITOR_TODO_COLLECTION)
+      .doc()
+      .set({
+        batchID,
+        pharmacyID: pharm.key,
+        data: pharm.values
+      })
```
auderephilip: Wow - glad this was so easy!

Q: I originally left out the await (see comment line 88) because forEach doesn't await anyway, and the await is inconsequential (i.e. no further code depends on it).  Would it potentially make more sense to either leave the awaits out, intentionally as before, or if we need the awaits, to `Promise.all()` instead?  I worry that the `async` and the `await` might imply to the casual reader that some form of promise-fulfillment is actually happening here, where in reality the `forEach` will will just move along without awaiting anything.

src/store/corestore.ts
```diff
@@ -129,8 +132,11 @@ async function loadAuditorTasks(): Promise<Task[]> {
       totalCost: d["Total med price covered by SPIDER"],
       claimedCost: d["Total reimbursement"],
       timestamp: new Date(d["YYYY"], d["MM"], d["DD"]).getTime(),
+    }));
+    return {
+      entries: patients,
       site: {
-        name: d["g3:B01 Pharmacy name"],
+        name: d[0]["g3:B01 Pharmacy name"],
         phone: "+254 867 5309"
```
goldjay: Shouldn't the phone be the same as the phone for the site?

balderdash: It would seem the CSVs we get from Maisha Meds don't include the pharmacy's phone number. Removing this number for now.

functions/src/index.ts
balderdash: TBH I just added the await to appease the linter, but I think it makes sense to `Promise.all()` instead.

balderdash:
Added a bit more info to the task list entry:
image

functions/src/index.ts
auderephilip: @balderdash : does `Promise.all()` actually work?  I might misunderstand the documentation... but I could have sworn that they wanted a promise returned from this function.

As long as it all holds together!

auderephilip:
Awesome addition to the task list entry!

functions/src/index.ts
balderdash: It should work. By making the function `async`, it now implicitly returns a promise.

PR #11, opened 2019-10-02
Sample claims when rendering claim details, not when importing the CSV

balderdash:
Per @goldjay's suggestion, keep all the entries when importing the CSV, and only sample them when rendering the details panel in the auditor view.


auderephilip:
Now that I think about it, it's probably pretty important to sample stably... when an admin or someone else refers to the task later, which items where sampled is probably going to be important to know (e.g. if you leave a note about which photos are blurry, it'll be helpful for the ops person to know which records had that).

Given PR #13 , it should work to even just write a `sampled: true` property into the Task itself, since #13 makes it so that tasks will be written back into Firestore at every step of the pipeline (such that other screens can load the same task later and know which one was shown).

If this all sounds too complicated, we could alternatively choose to sample on the server and write the `sampled` fact there (and into the data).  I guess one relevant question is what we think we'll do when moving over to AWS for the server (whether/how we'd differentiate sampled vs. unsampled records on the server side).

Adding @audererob and @billy-audere in case they have opinions on this.

rightparen:
If writing back state on view, consider writing the identity of the viewer.  This can help ensure that viewers don't double audit a given item.

Also with Firestore you would probably want to use a transaction or write the value to a subcollection to avoid conflicts.

auderephilip:
Oh, interesting, @rightparen .  I hadn't thought through the multi-viewer scenario yet... but we likely need to.

If we were staying on Firebase for sure (beyond auth), we could use their data-change subscription functionality to communicate who's viewing each item (the way that Google Sheets shows you who else is in the same cell).

I've added this item to our work-item-tracking doc.

balderdash:
Alright, I updated this PR to:
* Shuffle the samples once during the CSV import
* Show the first `N` samples on the auditor page, so everyone will see the same set
* Mark the specific samples that were shown with `reviewed: true` when the claim gets approved.

auderephilip:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FQh6NZWsFx1G1O%2Fgiphy.gif&token=awesome" data-gitmeme-token="awesome"></a>

PR #13, opened 2019-10-03
Implement full task flow

auderephilip:
This integrates an Operator tab, and also completes the entire workflow (all the way through to server edits).  Key changes:

* New Role of Operator (the person who calls pharmacies and follows up)
* New OperatorPanel
* Complete flow stored in Firebase.  If an Auditor or Payor declines, the task ends up at the Operator showing their latest notes.  If an Operator completes their work, the task moves to the Payor.  If a Payor marks payment complete, we store the completed task in the payor_complete_tasks folder in Firestore.
* Data model simplification:  since we operate on pharmacies as a basic unit, it seemed simpler just to have Tasks be ClaimTasks.  In other words, a task moves from person to person... but there aren't different types of tasks.  I've also moved Notes into the TaskChangeMetadata, which is debatable;  but the thought was that every step of the way, someone in a role should be able to add a Note to explain their action/change.

A task that was Declined to the operator:
image

And then when that Operator completed their follow-up:
image

And finally, after the payor marked that task Complete:
image

### DIFF
--- a/src/Screens/PayorPanel.tsx
+++ b/src/Screens/PayorPanel.tsx
@@ -6,11 +6,11 @@ import LabelWrapper from "../Components/LabelWrapper";
 import TextItem from "../Components/TextItem";
 import {
   ClaimEntry,
-  ClaimTask,
   Task,
-  tasksForRole,
-  UserRole,
-  ReimbursementTask
+  loadPayorTasks,
+  declinePayment,
+  getLatestTaskNote,
+  savePaymentCompletedTask
 } from "../store/corestore";
 import "./MainView.css";
 import "react-tabs/style/react-tabs.css";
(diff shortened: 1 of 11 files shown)


src/Screens/PayorPanel.tsx
```diff
@@ -91,18 +103,19 @@ class PayorPanel extends React.Component<Props, State> {
...
+        />
         <div className="mainview_button_row">
-          <Button
-            label="Payment Complete"
-            onClick={() => console.log("Payment complete")}
-          />
+          <Button label="Payment Complete" onClick={this._onPaymentComplete} />
         </div>
       </LabelWrapper>
     );
```
goldjay: On the Payor Panel, the details portion is a bit small. If you would make the div around TaskList width: 100% it would look nicer. Also, not sure sure how everyone feels about it, but I liked the tasks to the left of the details.

auderephilip: Oooh, yeah.  Sorry, I messed up the location of the details accidentally.  Will fix it, and also address the width business.  Thanks, @goldjay !

PR #14, opened 2019-10-03
Details Search Auditor View

goldjay:
* Search by values i.e. patient name, date or item
* Search by specific key using a ":" ex: "item:e" searches for all detail items with an item value that contains "e"
* Debounced input for less jarring UX

### DIFF
--- a/src/Screens/AuditorPanel.tsx
+++ b/src/Screens/AuditorPanel.tsx
@@ -4,18 +4,19 @@ import Button from "../Components/Button";
 import ImageRow from "../Components/ImageRow";
 import LabelTextInput from "../Components/LabelTextInput";
 import LabelWrapper from "../Components/LabelWrapper";
+import TaskList from "../Components/TaskList";
 import TextItem from "../Components/TextItem";
 import {
   ClaimEntry,
-  Task,
   declineAudit,
-  loadAuditorTasks,
   getLatestTaskNote,
-  saveAuditorApprovedTask
+  loadAuditorTasks,
+  saveAuditorApprovedTask,
+  Task
 } from "../store/corestore";
+import debounce from "../util/debounce";
+import { containsSearchTerm } from "../util/search";
 import "./MainView.css";
-import "react-tabs/style/react-tabs.css";
-import TaskList from "../Components/TaskList";
 
 const MIN_SAMPLE_FRACTION = 0.2;
 const MIN_SAMPLES = 1;
@@ -26,11 +27,13 @@ type State = {
   selectedTaskIndex: number;
   notes: string;
   numSamples: number;
+  searchPhrase: string;
 };
 
 class AuditorPanel extends React.Component<Props, State> {
   state: State = {
     tasks: [],
+    searchPhrase: "",
     selectedTaskIndex: -1,
     notes: "",
     numSamples: 0
@@ -70,7 +73,7 @@ class AuditorPanel extends React.Component<Props, State> {
     await saveAuditorApprovedTask(
       this.state.tasks[this.state.selectedTaskIndex],
       this.state.notes,
-      this.state.numSamples,
+      this.state.numSamples
     );
     this._removeSelectedTask();
   };
@@ -107,6 +110,7 @@ class AuditorPanel extends React.Component<Props, State> {
   };
 
   _renderClaimEntryDetails = (entry: ClaimEntry) => {
+    const { searchPhrase } = this.state;
     let patientProps = [];
     if (!!entry.patientAge) patientProps.push(entry.patientAge);
     if (!!entry.patientSex && entry.patientSex!.length > 0)
@@ -114,14 +118,23 @@ class AuditorPanel extends React.Component<Props, State> {
     const patientInfo =
       patientProps.length > 0 ? `(${patientProps.join(", ")})` : "";
 
+    const date = new Date(entry.timestamp).toLocaleDateString();
+    const patient = `${entry.patientFirstName} ${entry.patientLastName} ${patientInfo}`;
+    const random = Math.random() * 1000;
+    let checkEntry = { ...entry } as any;
+    checkEntry.date = date;
+    checkEntry.patient = patient;
+
+    if (!!searchPhrase && !containsSearchTerm(searchPhrase, checkEntry)) {
+      return null;
+    }
+
     return (
-      <LabelWrapper key={entry.patientID}>
-        <TextItem
-          data={{ Date: new Date(entry.timestamp).toLocaleDateString() }}
-        />
+      <LabelWrapper key={entry.patientID + random.toString()}>
+        <TextItem data={{ Date: date }} />
         <TextItem
           data={{
-            Patient: `${entry.patientFirstName} ${entry.patientLastName} ${patientInfo}`
+            Patient: patient
           }}
         />
         <TextItem data={{ Item: entry.item }} />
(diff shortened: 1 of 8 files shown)


src/Screens/AuditorPanel.tsx
```diff
@@ -107,21 +110,31 @@ class AuditorPanel extends React.Component<Props, State> {
...
     if (!!entry.patientAge) patientProps.push(entry.patientAge);
     if (!!entry.patientSex && entry.patientSex!.length > 0)
       patientProps.push(entry.patientSex);
     const patientInfo =
       patientProps.length > 0 ? `(${patientProps.join(", ")})` : "";
 
+    const date = new Date(entry.timestamp).toLocaleDateString();
+    const patient = `${entry.patientFirstName} ${entry.patientLastName} ${patientInfo}`;
+    const random = Math.random() * 1000;
+    let checkEntry = { ...entry } as any;
```
auderephilip: Any difference between this and `let checkEntry = Object.assign({}, entry)`?

Alternatively, if you're going to do this, might you be able to just:
```
  let cE = {
    ...entry,
    date,
    patient
};
```

src/Screens/AuditorPanel.tsx
```diff
@@ -107,21 +110,31 @@ class AuditorPanel extends React.Component<Props, State> {
...
+    if (!!searchPhrase && !containsSearchTerm(searchPhrase, checkEntry)) {
+      return null;
+    }
+
     return (
-      <LabelWrapper key={entry.patientID}>
-        <TextItem
-          data={{ Date: new Date(entry.timestamp).toLocaleDateString() }}
-        />
+      <LabelWrapper key={entry.patientID + random.toString()}>
```
auderephilip: Not super-positive I understand this.  By doing this, it guarantees that react always treats this object as different every time, right?  Do we actually want that?  Might it make sense to bind the uniqueness of the object more to whatever really reflects the change you want to make unique (e.g. is it Patient + Search Term?).

src/Screens/AuditorPanel.tsx
```diff
@@ -130,13 +143,29 @@ class AuditorPanel extends React.Component<Props, State> {
     );
   };
 
+  _setSearchPhrase = debounce((input: string) => {
+    this.setState({ searchPhrase: input });
```
auderephilip: In the past, I've seen this done differently, so that the text itself isn't necessarily stored in state repeatedly, but only stored whenever a delay-launched background search happens (i.e. tying the saved state to whatever's actually been searched/filtered and displayed).

IIRC, there are some classic concerns with react loops if textbox onChange events get mapped into state that then renders the textbox differently (it causes flickering).  If this variable doesn't change rendering, it might be better to store it in a private member (e.g. this._searchPhrase);  then you wouldn't need to debounce (or if the search is too compute-intensive, you can setTimeout() on the search function to only run it when rapid typing has stopped).

src/Screens/AuditorPanel.tsx
```diff
@@ -167,7 +199,7 @@ class AuditorPanel extends React.Component<Props, State> {
           renderItem={this._renderTaskListClaim}
           className="mainview_tasklist"
         />
-        <div>
+        <div style={{ width: "100%" }}>
```
auderephilip: I'm not sure what's going on with this.  Worked fine for me without the 100%... but needed it in PayorPanel.

src/Screens/OperatorPanel.tsx
```diff
@@ -4,19 +4,20 @@ import Button from "../Components/Button";
...
+  saveOperatorCompletedTask,
+  Task
 } from "../store/corestore";
 import "./MainView.css";
-import "react-tabs/style/react-tabs.css";
-import TaskList from "../Components/TaskList";
 
-type Props = {};
+type Props = {
+  searchPhrase?: string;
```
auderephilip: Not sure I understand why this was added to props if nothing else in the file was changed...

src/util/search.ts
```diff
@@ -0,0 +1,33 @@
+export const containsSearchTerm = (
+  searchPhrase: string,
+  entry: any
+): boolean => {
+  if (searchPhrase.length > 0) {
```
auderephilip: nit: personal preference when a function is long is to not have the exceptional "else" case be a tiny thing at the end.  The inverted phrasing is often easier to really understand quickly:

```
if (searchPhrase.length === 0) {
  return false;
}

// ... rest of func
```

src/util/search.ts
```diff
@@ -0,0 +1,33 @@
...
+    const keys = Object.keys(entry);
+
+    let searchKey;
+    if (searchPhrase.includes(":")) {
```
auderephilip: Neat.  A fun approach, allowing for key:value searches.

src/util/search.ts
```diff
@@ -0,0 +1,33 @@
...
+      const searchArr = searchPhrase.split(":");
+      searchKey = searchArr[0];
+      searchPhrase = searchArr[1];
+    }
+    for (let i = 0; i < keys.length; i++) {
+      const value = entry[keys[i]];
+      if (
+        ((!!searchKey &&
+          keys[i].toLowerCase().includes(searchKey.toLowerCase())) ||
```
auderephilip: In the case of key:value searches, might it instead make most sense to require an exact key match?  Maybe... maybe not.  Worth thinking about.

src/util/search.ts
```diff
@@ -0,0 +1,33 @@
...
+  entry: any
+): boolean => {
+  if (searchPhrase.length > 0) {
+    const keys = Object.keys(entry);
+
+    let searchKey;
+    if (searchPhrase.includes(":")) {
+      const searchArr = searchPhrase.split(":");
+      searchKey = searchArr[0];
+      searchPhrase = searchArr[1];
```
auderephilip: Consider `trim`.

src/util/search.ts
```diff
@@ -0,0 +1,33 @@
+export const containsSearchTerm = (
+  searchPhrase: string,
```
auderephilip: Consider `trim`, especially in how it impacts substring results.

src/Screens/AuditorPanel.tsx
goldjay: The spread operator and Object.assign are just about the same, but you can mutate the object while copying it. I'll change it for consistency.

src/Screens/OperatorPanel.tsx
goldjay: Left over / added back in from when I had the search bar in main view. I'll take it out.

src/Screens/AuditorPanel.tsx
goldjay: This was to mitigate some of the weirdness that came with typing into the search box. If nothing was found, the details panel would shrink horizontally.

src/Screens/AuditorPanel.tsx
goldjay: I was getting super weird behavior with the details since there were duplicate patientIDs (Duplicated details when there was more than one claim) and this was my way to not have a collision while I was working on it. I'll change it to entry.patientId + patient.

src/util/search.ts
goldjay: My thought is if I were using this system for a while it would be nice to not have to type the whole key in when doing a search. Right now there aren't many keys/values so you could use the first letters like "i:c" to find any condoms... it's a feature?

auderephilip: Sounds good!  If so, recommend `trim()` to make sure you catch all the substrings.

src/Screens/AuditorPanel.tsx
goldjay: Maybe I am misunderstanding, but SetState is only called once the input has been inactive for half a second.

auderephilip: Oh, I get it.  Sorry, misunderstood the original design.  You're essentially debouncing the actual UI refresh;  makes sense!  Nothing to see here!

src/Screens/AuditorPanel.tsx
```diff
@@ -107,21 +110,29 @@ class AuditorPanel extends React.Component<Props, State> {
...
     if (!!entry.patientAge) patientProps.push(entry.patientAge);
     if (!!entry.patientSex && entry.patientSex!.length > 0)
       patientProps.push(entry.patientSex);
     const patientInfo =
       patientProps.length > 0 ? `(${patientProps.join(", ")})` : "";
 
+    const date = new Date(entry.timestamp).toLocaleDateString();
+    const patient = `${entry.patientFirstName} ${entry.patientLastName} ${patientInfo}`;
+
+    let checkEntry = Object.assign({}, entry, date, patient);
```
auderephilip: Neat.  TIL.  Didn't realize you could just add additional fields this way.

src/util/debounce.ts
```diff
@@ -0,0 +1,11 @@
+// FROM: https://gist.github.com/ca0v/73a31f57b397606c9813472f7493a940
+export default function debounce<F extends (...params: any[]) => void>(
```
auderephilip: So surprised this wasn't part of a standard, common library, even in React.  Really useful for UI code.

Seems `lodash` and other libraries would be too heavyweight for just getting this.

PR #15, opened 2019-10-04
Implements payments via Africa's Talking

auderephilip:
The following changes are included:
* A cloud function that calls the REST APIs for payments
* A JSON file with the actual API key which needs to be installed on every engineer's machine (stored in LastPass)
* An enhanced Payor tab with the new functionality
* Added ability to disable a Button

image

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -1,6 +1,8 @@
 import * as functions from "firebase-functions";
 import * as admin from "firebase-admin";
-import * as csvtojson from "csvtojson";
+import csvtojson from "csvtojson";
+import axios, { AxiosResponse } from "axios";
+import africasTalkingOptions from "./africas-talking-options.json";
 
 // You're going to need this file on your local machine.  It's stored in our
 // team's LastPass ServerInfrastructure section.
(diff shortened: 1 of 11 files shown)


src/Components/Button.css
```diff
@@ -9,11 +9,15 @@
...
+.button_container.enabled:hover {
   cursor: pointer;
 }
 
-.button_container:active {
+.button_container.enabled:active {
   background-color: grey;
   color: white;
 }
+
```
audererob: I'm not 100% sure this would work, but for input disabling you might be able to do something like:

.button_container.disabled {
  pointer-events: none
  tabIndex: undefined
}

src/Components/Button.css
```diff
@@ -9,11 +9,15 @@
   margin: 5px;
 }
 
-.button_container:hover {
+.button_container.enabled:hover {
```
goldjay: You can stack pseudo-classes in css i.e .button_container:enabled:hover. I know I'm the one that originally made it a div (I don't like having to override all of the default button styling), but it may be cleaner to make the button an html button so we could used the "disabled" property.

functions/src/index.ts
```diff
@@ -27,37 +29,72 @@ type CallResult =
...
+    if (!hasRole(context, UserRole.PAYOR)) {
+      throw Error(`User ${context.auth && context.auth.uid} isn't a Payor`);
+    }
+    if (!data || !data.recipients) {
+      throw Error("No recipients specified");
+    }
+
+    const endpoint = isProductionProject()
+      ? "https://payments.africastalking.com/mobile/b2c/request"
+      : "https://payments.sandbox.africastalking.com/mobile/b2c/request";
```
balderdash: Consider setting this URL in the africa's talking config file as well, rather than selecting based on the environment. The implementation of `isProductionProject` also seems kinda fragile.

auderephilip:
Great points, everyone.  Will make a bunch of changes and pop back.

functions/src/index.ts
auderephilip: Great call.  Brain fart on missing that.  Much better implementation - thanks for suggesting, @balderdash !

src/Components/Button.css
auderephilip: Ah, right then.  Ok.  Will try the default button business, and also implement the stacking.

src/Components/Button.tsx
```diff
@@ -3,19 +3,34 @@ import "./Button.css";
 
 interface Props {
   className?: string;
+  disabledClassName?: string;
```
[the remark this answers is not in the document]
auderephilip: Not sure... my thinking was that if we want to support `className` normally, we'd probably want the caller to also be able to specify what things look like disabled.  But perhaps it'd be cleaner just to apply some sort of greying transform when disabled;  I'll dork around with it.

src/store/corestore.ts
```diff
@@ -293,6 +308,18 @@ export async function setRoles(email: string, roles: UserRole[]) {
   }
 }
 
+export async function issuePayments(recipients: PaymentRecipient[]) {
+  // You need a big timeout on this (e.g. 300,000 msec), because Africa's
+  // Talking can sometimes go 1-2 minutes before responding back with a result.
```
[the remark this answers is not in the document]
auderephilip: Right?!   Not a joke.  I've been active inside their Slack instance trying to drum up some interest to talk  about this issue.

src/Components/Button.css
auderephilip: Oooh.  TIL.  Will attempt this.  Thanks, @audererob .

auderephilip: Ended up using standard `<button>`, as suggested elsewhere, and it made this all a bit cleaner.

PR #1, opened 2019-09-25
Initial incentives site checkin

auderephilip:
This was pretty much ported from a previous commit (https://github.com/AudereNow/audere/pull/192).  Gets a basic React site up and running with Firebase auth, and updates automatically (hot-loads) when you use 'yarn start' to run the code.

PR #16, opened 2019-10-04
Dropdown Menu + TaskList Hover Behavior

goldjay:
* Configurable Dropdown Menu
* Added CSS to TaskList to change color on hover and have the pointer cursor
* Logout Functionality

Q: HamburgerIcon is in HTML/CSS. Should it be in /assets? 

Dropdown + TaskList Hover

### DIFF
--- a/src/Components/Dropdown.tsx
+++ b/src/Components/Dropdown.tsx
@@ -1,0 +1,65 @@
+import React, { PureComponent } from "react";
+import "./Dropdown.css";
+import HamburgerIcon from "./HamburgerIcon";
+
+interface dropdownConfig {
+  label?: string;
+  onSelect?: () => void;
+}
+
+interface Props {
+  config: dropdownConfig[];
+  labelURI?: string;
+}
+
+interface State {
+  open: boolean;
+}
+
+export default class Dropdown extends PureComponent<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = { open: false };
+  }
+
+  _toggleDropdown = () => {
+    this.setState({ open: !this.state.open });
+  };
+
+  render() {
+    if (this.props.config.length === 0) {
+      return null;
+    }
+    const { open } = this.state;
+    const { labelURI } = this.props;
+
+    return (
+      <div
+        className="dropdown_container"
+        onMouseEnter={this._toggleDropdown}
+        onMouseLeave={this._toggleDropdown}
+      >
+        {!!labelURI ? (
+          <img src={labelURI} alt={labelURI} />
+        ) : (
+          <HamburgerIcon active={open} height={20} width={20} />
+        )}
+        {!!open && (
+          <div className="dropdown_content">
+            {this.props.config.map(item => {
+              return (
+                <div
+                  key={item.label}
+                  className="dropdown_item"
+                  onClick={item.onSelect}
+                >
+                  <span>{item.label}</span>
+                </div>
+              );
+            })}
+          </div>
+        )}
+      </div>
+    );
+  }
+}
--- a/src/Components/Dropdown.tsx
+++ b/src/Components/Dropdown.tsx
@@ -8,7 +8,6 @@ interface dropdownConfig {
 }
 
 interface Props {
-  config: dropdownConfig[];
   labelURI?: string;
 }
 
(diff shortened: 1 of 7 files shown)


src/Components/Dropdown.tsx
```diff
@@ -0,0 +1,65 @@
...
+      return null;
+    }
+    const { open } = this.state;
+    const { labelURI } = this.props;
+
+    return (
+      <div
+        className="dropdown_container"
+        onMouseEnter={this._toggleDropdown}
+        onMouseLeave={this._toggleDropdown}
```
auderephilip: Are there ever any cases (e.g. with event bubbling, etc) where a MouseEnter might not be coupled with a MouseLeave?  Like if I hover the mouse over the same pixel, but in another app that's covering my browser, and then switch apps such that the mouse is right over the hover menu?

I'm not positive if such cases could (ever) happen, or if browsers guarantee that under no conditions would these conditions be unpaired.  But it might be safer to have `this._showDropdown` and `this._hideDropdown` for avoidance of doubt.

src/Components/HamburgerIcon.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+class HamburgerIcon extends PureComponent<Props> {
+  render() {
+    const { active, height, width } = this.props;
+    const barColorStyle = { backgroundColor: active ? "white" : "black" };
+    return (
+      <div
+        className="hamburger_container"
+        style={{ width: `${width}px`, height: `${height}px` }}
```
auderephilip: tiny: thoughts on this approach vs. taking a `className` property?

src/Components/TopBar.tsx
```diff
@@ -23,6 +24,14 @@ class TopBar extends React.Component {
     this.setState({ roles });
   }
 
+  _handleLogout = async () => {
+    try {
+      await firebase.auth().signOut();
+    } catch (e) {
+      console.log(`Error logging out: ${e}`);
```
auderephilip: tiny: I've been a bit ambivalent, having turned some of these logs into alerts instead (thinking it might be helpful for the user to actually know that something has gone bad).

auderephilip: Q: do we already sniff for auth status changes, and show the auth screen after logout?  I suspect we do.

src/Components/TopBar.tsx
```diff
@@ -63,14 +72,30 @@ class TopBar extends React.Component {
...
         <img className="topbar_logo" src={logo} alt="logo" />
         <div className="topbar_user">
           {firebase.auth().currentUser!.displayName}
         </div>
-        {uploadButton}
-        {uploader}
+        <div
+          style={{
+            display: "flex",
+            flexDirection: "row"
```
auderephilip: I think we should make this into a css property to be consistent with the approach several lines above it -- so that styling is more consistently applied in the app (i.e. mostly through css in files, not inline).

src/Components/Dropdown.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { PureComponent } from "react";
+import "./Dropdown.css";
+import HamburgerIcon from "./HamburgerIcon";
+
+interface dropdownConfig {
+  label?: string;
```
auderephilip: Thoughts on this instead being `item: JSX.Element` or `item: React.Component<any>` instead?

I ask, because it'd be awesome to move the Upload functionality into the menu.  And I wonder if it'd be simple (or perhaps weird and odd) to have the upload file-selection functionality inline.  Perhaps it's too weird.

But taking a component allows you to have menu items with icons, etc etc.

Might be overkill.  Just a thought for you.

goldjay: Do you mean passing components as children to a menu wrapper? That sounds great!

src/Components/Dropdown.tsx
goldjay: That's a good point. I didn't think about a situation where the mouse would enter, but not leave or vice versa. I suppose you could mouse-enter and change tabs/programs using hotkeys. I'll change it to show and hide just in case.

src/Components/HamburgerIcon.tsx
goldjay: Seeing as we're already passing classNames elsewhere, it would be the way to go for consistency. One reason I'm hesitant to pass classNames is we are reliant on typing the className properly with no error detection when passing a non-existant/misspelled class and I've been burned by my fat fingers before. Passing properties seems like the more "Reacty" way to do it and Typescript checks that it's valid. Also, on the CSS hierarchy tree inline styles are #1, meaning they will override any existing styles.

PR #21, opened 2019-10-08
Adding task expand to auditor panel, autoselect to all panels, and fix search crash

audererob:
This adds an "expand/collapse" button at the end of the sampled tasks in the auditor panel to show all of them, as well as auto-selecting the first item in the tasklist when switching between panels. There's also a small tweak to search logic to avoid a crash.


src/util/search.ts
```diff
@@ -2,29 +2,29 @@ export const containsSearchTerm = (
...
-        keys[i].toLowerCase().includes(searchKey.toLowerCase())) ||
+      ((!!searchKey && keys[i].toLowerCase().includes(searchKey)) ||
         !searchKey) &&
       !!value &&
       value
         .toString()
+        .trim()
         .toLowerCase()
-        .includes(searchPhrase.toLowerCase())
+        .includes(searchPhrase)
```
auderephilip: Great edit.  Good call.

src/Screens/AuditorPanel.tsx
```diff
@@ -28,6 +28,7 @@ type State = {
   notes: string;
   numSamples: number;
   searchPhrase: string;
+  showAllTasks: boolean;
```
auderephilip: I was a little confused about this until I figured out that this, semantically, actually is `showAllEntries`.  Could we rename things around a bit so that it's the case?

PR #22, opened 2019-10-08
Adding animated image zoom w/modal behavior

audererob:
When hovering over an image, clicking it will zoom in to it with an animation and the background will dim. Also tweaked a bit of text.


package.json
```diff
@@ -13,6 +13,7 @@
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
     "react-firebaseui": "^4.0.0",
+    "react-pose": "^4.0.9",
```
auderephilip: TIL

src/Components/ImageRow.tsx
```diff
@@ -11,8 +12,9 @@ const ImageRow = (props: Props) => {
       {props.imageURLs.length > 0 &&
         props.imageURLs.map((url, index) => {
           return (
-            <img
-              className="imagerow_image"
+            <ZoomableImage
+              width="20%"
+              height="20%"
```
auderephilip: Really useful.

src/Components/ZoomableImage.tsx
```diff
@@ -0,0 +1,80 @@
...
+  width?: string | number;
+  height?: string | number;
+  [key: string]: any;
+}
+
+interface State {
+  isZoomed: boolean;
+}
+
+class ZoomableImage extends Component<Props> {
```
auderephilip: Might there be a popular React component that does this?

src/Screens/OperatorPanel.tsx
```diff
@@ -121,7 +121,7 @@ class OperatorPanel extends React.Component<Props, State> {
 
   _renderClaimDetails = (task: Task) => {
     return (
-      <LabelWrapper label="DETAILS VIEW">
+      <LabelWrapper label="DETAILS">
```
auderephilip: Now that I see all the changes in Panels that we keep making in parallel, I wonder if my original factoring is just really bad.  Perhaps neither here nor there for now... but I do wonder whether there's a better common factoring of all Panels so that diffs are less repetitive.  My fault.  (No need to correct it in this PR;  just remarking).

src/Components/ZoomableImage.tsx
audererob: Great point, there's a pretty popular component here - https://github.com/rpearce/react-medium-image-zoom. We can certainly look at switching over - I was going to work on incorporating text into the zoom which may or may not be easier with our own component.

auderephilip: Totally fine with us rolling our own, especially if we're going to add functionality.  And especially if it's a simple component like this one.  I'm supportive!

PR #25, opened 2019-10-10
Notes Audit Trail for Operator and Payor

goldjay:
Payor
Screen Shot 2019-10-10 at 3 04 33 PM
Auditor
Screen Shot 2019-10-10 at 3 04 21 PM

### DIFF
--- a/src/Components/NotesAudit.css
+++ b/src/Components/NotesAudit.css
@@ -1,0 +1,4 @@
+.notesaudit_row {
+  margin: 10px;
+  overflow-wrap: break-word;
+}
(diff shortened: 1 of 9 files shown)


audererob:
Is it possible to refactor the notes rendering into a common method or simple component? Also, I *think* we'd want to remove the Notes field's default value getting set to getLatestTaskNote? Unless I'm misunderstanding the intent of that.

goldjay:
@audererob Yeah, I thought about that when I was copy/pasting the code from one panel to another, but I wasn't sure if it was too small or not. I'll make it into a simple component. @auderephilip Correct me if I'm wrong, but it seems like we can remove the note field's default value and getLatestTaskNote.

auderephilip:
> @auderephilip Correct me if I'm wrong, but it seems like we can remove the note field's default value and getLatestTaskNote.

Indeed, good to remove.  Thanks!

PR #27, opened 2019-10-11
Adding payment decline and claim rejection, as well as filtering of tasks

audererob:
This change adds "Decline Payment" to the Payor panel, as well as "Reject" to the Operator panel. Additionally, filtering tasks by "Todo", "Approved" and "Rejected" is in the Auditor panel.


balderdash:
Would it be clearer to put declined/rejected and completed claims into their own tabs, and show people with the `Auditor` role all 3 tabs?

Sorry if this was already discussed while I was out. The code LGTM.

audererob:
I like that idea - I'll work on a follow-up change.

PR #29, opened 2019-10-14
Fix CSV uploads

balderdash:
CSV uploads weren't working because the menu disappeared lost focus while you were selecting the file to upload. Pin it in place while you're selecting the file.


balderdash:
Oh, sorry about the duplicate effort!

PR #30, opened 2019-10-14
Upload CSV Fix

goldjay:
Two things:
1. CSV upload wasn't working because it was in the menu. The input gets removed on mouse-out / dropdown closing so it isn't around to register the input on change event on file selection and doesn't upload the CSV.
2. Found a nice way to use SVG as react components instead of img src! This means we can take advantage of CSV stuff like changing the fill color on hover.


auderephilip:
Whoa.  TIL about #2.  Fascinating.

PR #34, opened 2019-10-15
Date Search for Pharmacies + Search Clear

goldjay:
* Date filter
    - Can enter a start date, end date or both to get a complete range

* Clear filter

Question: All of this passing functions and variables down several components feels strange. Is now a good time to integrate Redux?


src/Components/LabelWrapper.tsx
```diff
@@ -4,13 +4,21 @@ import "./LabelWrapper.css";
...
     return (
       <div className={`labelwrapper_container ${this.props.className}`}>
-        <div className="labelwrapper_header">{this.props.label}</div>
+        <div className="labelwrapper_header">
+          {this.props.label}
+          {!!this.props.renderLabelItems && (
+            <div className="labelwrapper_labelitems_container">
+              {this.props.renderLabelItems()}
+            </div>
+          )}
```
auderephilip: Would instead using `this.props.children` accomplish the intent of this?  Like would we be able to not have `renderLabelItems` at all, but accept `<LabelWrapper>` components with child components?

src/Screens/AuditorPanel.tsx
```diff
@@ -251,8 +271,8 @@ class AuditorPanel extends React.Component<Props, State> {
           task.entries
             .slice(this.state.numSamples, task.entries.length)
             .map(this._renderClaimEntryDetails)}
-        {task.changes.map(change => {
-          return <NotesAudit change={change} />;
+        {task.changes.map((change, index) => {
+          return <NotesAudit key={change.by + index} change={change} />;
```
auderephilip: Thanks for fixing this.  I hit this in the console just this evening.

src/Screens/AuditorPanel.tsx
```diff
@@ -283,23 +303,34 @@ class AuditorPanel extends React.Component<Props, State> {
...
+        return (
+          (containsSearchTerm(searchTerm, task.site) ||
+            task.entries.some(entry => {
+              return containsSearchTerm(searchTerm, entry);
+            })) &&
+          task.entries.some(entry => {
+            return withinDateRange(dateRange, entry);
+          })
+        );
+      }
```
auderephilip: Not positive I understand the logic correctly, but could the big `if` be removed entirely, and the `else` clause changed to effectively say:

```
  return (
    searchTerm === "" ||
    contains(blah) ||
    entries.some(contains())
) && entries.some(withinDate())
```

In other words, was your intent to modify the original way of filtering to always require that a date range be met?  In which case, this clause feels like it'd be simpler if we just added the date condition with an `&&` at the end.

src/Components/LabelWrapper.tsx
goldjay: Unfortunately we're already using this.props.children to render the TaskList or details and it would be a little tricky to distinguish between children to be rendered as a labelItem or as an inner child.

src/Screens/AuditorPanel.tsx
goldjay: Nice! I made a small change to containsSearchTerm and it works with the more straightforward logic. My original intention was to preserve the edge-case catch outside of the search functions since that's the way it was, but I think it makes more sense this way.

PR #36, opened 2019-10-16
FEV-1343 fix month off by one

tpaik18:
https://auderenow.atlassian.net/browse/FEV-1343


tpaik18:
Oops, doesn't look like I did this right.

tpaik18:
Looking at it some more, I don't see any better way to fix this.

src/store/corestore.ts
```diff
@@ -325,7 +325,7 @@ export async function loadAuditorTasks(): Promise<Task[]> {
         item: d["Type received"],
         totalCost: parseFloat(d["Total med price covered by SPIDER"]),
         claimedCost: parseFloat(d["Total reimbursement"]),
-        timestamp: new Date(d["YYYY"], d["MM"], d["DD"]).getTime()
+        timestamp: new Date(d["YYYY"], d["MM"] - 1, d["DD"]).getTime()
```
auderephilip: Ack.  Sorry.  Didn't even think about this.

PR #38, opened 2019-10-17
Use a more generic TaskPanel to render Payor and Operator tabs

balderdash:
I'm trying to do two things with this PR:
* Reduce code duplication across tabs
* Move to a more config-driven UI for flows

To those ends, I created a config file under `src/store` that defines the tabs that show up in our app, and a "TaskPanel" component that is meant to render any workflow step based on that config.

For a normal workflow tab, the configuration lists the collection to load tasks from, and the components to use to render the task list item and the details panel. I also plan to make the accept/decline buttons configurable (they mostly do the same thing: save notes and move the task to a different collection), but for now they're still just hardcoded into the respective "details" components.

For non-workflow tabs like the Admin tab, the config can also just accept a component to render the entire tab. I'm also using that for the auditor tab for now, since it has some additional features, but I'll coax that into a set of normal workflow tabs in a followup PR.

### DIFF
--- a/src/Screens/MainView.tsx
+++ b/src/Screens/MainView.tsx
@@ -1,24 +1,39 @@
 import React from "react";
 import { Tab, TabList, TabPanel, Tabs } from "react-tabs";
 import "react-tabs/style/react-tabs.css";
-import { UserRole, userRoles } from "../store/corestore";
+import { UserRole, userRoles, Task } from "../store/corestore";
 import AdminPanel from "./AdminPanel";
 import AuditorPanel from "./AuditorPanel";
 import MainChrome from "./MainChrome";
 import "./MainView.css";
 import OperatorPanel from "./OperatorPanel";
-import PayorPanel from "./PayorPanel";
+import { PayorItem, PayorDetails } from "./PayorPanel";
+import { isCustomPanel, defaultConfig } from "../store/config";
+import TaskPanel from "./TaskPanel";
 
 type Props = {};
 type State = {
   roles: UserRole[];
 };
 
-const RoleToPanelMap = {
-  Auditor: AuditorPanel,
-  Payor: PayorPanel,
-  Operator: OperatorPanel,
-  Admin: AdminPanel
+const PanelComponents: {
+  [key: string]: React.ComponentClass<{}>;
+} = {
+  Admin: AdminPanel,
+  AuditTask: AuditorPanel,
+  OperatorTask: OperatorPanel
+};
+
+const ItemComponents: {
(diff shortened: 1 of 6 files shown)


src/Screens/MainView.tsx
```diff
@@ -1,24 +1,40 @@
...
   roles: UserRole[];
 };
 
-const RoleToPanelMap = {
-  Auditor: AuditorPanel,
-  Payor: PayorPanel,
-  Operator: OperatorPanel,
-  Admin: AdminPanel
+const PanelComponents: {
+  [key: string]: React.ComponentClass<{}>;
```
auderephilip: TIL you could do this.  Neat.

src/Screens/TaskPanel.tsx
```diff
@@ -0,0 +1,84 @@
...
+        task => task.id === this.state.selectedTaskId
+      );
+      if (newIndex === -1) {
+        newIndex = Math.min(this.state.selectedTaskIndex, tasks.length - 1);
+      }
+      this.setState({
+        selectedTaskIndex: newIndex,
+        selectedTaskId: tasks[newIndex].id
+      });
+    });
```
auderephilip: Would prefer handlers this long to be named semantically as a function (e.g. `_onTasksChanged`).

src/Screens/PayorPanel.tsx
```diff
@@ -196,29 +152,28 @@ class PayorPanel extends React.Component<Props, State> {
...
-        <div style={{ width: "100%" }}>
-          {selectedTaskIndex >= 0 &&
-            this._renderReimbursementDetails(
-              this.state.tasks[selectedTaskIndex]
-            )}
+      <div className={previewName}>
+        <div className="mainview_preview_header">
+          <span>{this.props.task.site.name}</span>
         </div>
+        <div>{"Total Reimbursement: " + claimsTotal.toFixed(2) + " KSh"}</div>
```
tpaik18: Would be best if this used the formatCurrency() function, thanks.

src/store/corestore.ts
```diff
@@ -278,12 +278,16 @@ export function getBestUserName(): string {
...
-    .collection(PAYOR_TASK_COLLECTION)
-    .get();
-  return taskSnapshot.docs.map(doc => (doc.data() as unknown) as Task);
+    .collection(taskCollection)
+    .onSnapshot(snapshot =>
+      callback(snapshot.docs.map(doc => (doc.data() as unknown) as Task))
+    );
 }
 
 export async function loadOperatorTasks(): Promise<Task[]> {
```
audererob: I think you can get rid of this now.

PR #39, opened 2019-10-17
WIP: UI test of login with google

smeds1:
This PR creates an automated test to log into Jasmine. It uses my Audere email and password, which I stored as an environmental variable on my local machine. If we decide to develop the automation, I suspect we would want to change this to a different account and use a .env file. 

@balderdash, would there be any way to run UI tests locally with a different csv file than staging is using? I could set up a lot of tests to approve items, issue payment, etc. but it would only work well if I knew what data was supposed to be present and could guarantee that it always started in the same configuration.

### DIFF
--- a/test/ui.test.tsx
+++ b/test/ui.test.tsx
@@ -1,0 +1,37 @@
+// Copyright (c) 2019 by Audere
+//
+// Use of this source code is governed by an MIT-style license that
+// can be found in the LICENSE file distributed with this file.
+
+require('chromedriver');
+const assert = require('assert');
+const {Builder, Key, By, until} = require('selenium-webdriver');
+const email_password = process.env.GMAIL_PASS;
+
+describe('Login to Jasmine', function () {
+    let driver;
+    beforeAll(async function() {
+        driver = await new Builder().forBrowser('chrome').build();
+    });
+
+    test('Login with google', async function() {
+
+        await driver.get('https://flows-app-staging.firebaseapp.com/');
+
+        let title = await driver.getTitle();
+        assert.equal(title, 'React App'); //someone should make a better title
+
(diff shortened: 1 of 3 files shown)


test/ui.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+    beforeAll(async function() {
+        driver = await new Builder().forBrowser('chrome').build();
+    });
+
+    test('Login with google', async function() {
+
+        await driver.get('https://flows-app-staging.firebaseapp.com/');
+
+        let title = await driver.getTitle();
+        assert.equal(title, 'React App'); //someone should make a better title
```
tpaik18: I totally agree!

test/ui.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+        await driver.findElement(By.className('firebaseui-idp-google')).click();
+        await driver.wait(until.elementLocated(By.id('identifierId')), 10000);
+        await driver.findElement(By.id('identifierId')).sendKeys('sam@auderenow.org');
```
tpaik18: I think better to have the email address be an env variable also, so anyone else can run.

balderdash:
I looked into running flow against a local emulator, but it seems like there isn't good support for emulating the cloud storage service we use. While we could work around that, it might be easier to just spin up a separate test project on firebase, and use that for UI testing. We can have a script or a test-only endpoint that wipes out the existing tasks so you can start from scratch each time.

While it would be nice to have this level of testing eventually, I'm not sure it's worth investing too much time into at the moment, since the project is still in flux.

smeds1:
@balderdash Thanks for investigating. I agree that it doesn't make sense spending a ton of time setting up a rich test environment at the moment. I'll take the PR down for now and revisit later if we decide to invest in automated UI testing.

PR #44, opened 2019-10-18
FEV-1346: Added date to payor tab

smeds1:
I added a date column to the payor tab. While doing so, I noticed that the items don't seem to be sorted in any particular order. Should we sort by date?

image


smeds1:
Now sorted by date:

image

PR #48, opened 2019-10-18
Download CSV

goldjay:
* Decided to use json-2-csv instead of the function I wrote because it has some schema detection beyond looking at the keys of the first row.
* Also adjusted the margin/styling of the notes field


src/Screens/AuditorPanel.tsx
```diff
@@ -435,6 +436,44 @@ class AuditorPanel extends React.Component<Props, State> {
...
+        if (!csv || err) {
+          alert("Something went wrong when trying to download your csv");
+        }
+
+        const dataString = "data:text/csv;charset=utf-8," + csv;
+        const encodedURI = encodeURI(dataString);
+        const link = document.createElement("a");
+        link.setAttribute("href", encodedURI);
+        link.setAttribute("download", `${fileName}.csv`);
+        link.click();
```
auderephilip: Whoa.  I had no idea this is how a download is done.  Weird!

PR #50, opened 2019-10-21
Entering a Note on One Task Makes It Appear on All 

goldjay:
[FEV-1338](https://auderenow.atlassian.net/browse/FEV-1338)

* "defaultValue" on textarea doesn't update on prop change, but "value" does
* Updates local notes state when the task has changed

### DIFF
--- a/src/Components/LabelTextInput.tsx
+++ b/src/Components/LabelTextInput.tsx
@@ -22,7 +22,7 @@ class LabelTextInput extends Component<Props> {
         )}
         <textarea
           className="labeltextinput_input"
-          defaultValue={this.props.defaultValue}
+          value={this.props.defaultValue}
           onChange={this._onTextChange}
         />
       </div>
--- a/src/Screens/OperatorPanel.tsx
+++ b/src/Screens/OperatorPanel.tsx
@@ -26,7 +26,7 @@ export class OperatorDetails extends React.Component<Props, State> {
     notes: ""
   };
 
-  componentWillReceiveProps(nextProps: Props) {
+  componentDidUpdate(nextProps: Props) {
     if (nextProps.task !== this.props.task) {
       this.setState({ notes: "" });
     }
--- a/src/Screens/OperatorPanel.tsx
+++ b/src/Screens/OperatorPanel.tsx
@@ -26,6 +26,12 @@ export class OperatorDetails extends React.Component<Props, State> {
     notes: ""
   };
 
+  componentWillReceiveProps(nextProps: Props) {
+    if (nextProps.task !== this.props.task) {
+      this.setState({ notes: "" });
+    }
+  }
+
   _onNotesChanged = (notes: string) => {
     this.setState({ notes });
   };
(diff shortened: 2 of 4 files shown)


src/Screens/OperatorPanel.tsx
```diff
@@ -26,6 +26,12 @@ export class OperatorDetails extends React.Component<Props, State> {
     notes: ""
   };
 
+  componentDidUpdate(nextProps: Props) {
```
auderephilip: My understanding is that this is now deprecated, and that the recommended practice is https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops...

src/Components/LabelTextInput.tsx
```diff
@@ -22,7 +22,7 @@ class LabelTextInput extends Component<Props> {
         )}
         <textarea
           className="labeltextinput_input"
-          defaultValue={this.props.defaultValue}
+          value={this.props.defaultValue}
```
auderephilip: Is this actually safe?  Like if React updates the component for any of its reasons, won't `defaultValue` overwrite whatever current value the user has put into the control?

The pattern I've seen people do in the past is to have `state` track a `currentValue`, render the textarea with `value={currentValue}`, have `_onTextChange` update `currentValue`, and `getDerivedStateFromProps()` update `defaultValue` and `currentValue` when `defaultValue` changes.

src/Screens/OperatorPanel.tsx
goldjay: I used componentDidUpdate as suggested after getting a console depreciation warning for using "componentDidReceiveProps". 🤔 Looking at the docs, it looks like getDerivedStateFromProps gets called before componentDidUpdate and before the render so I'll make the change.

src/Components/LabelTextInput.tsx
goldjay: If React updates the component, the notes field will have whatever is currently in the notes state of its panel. The problem I ran into was that defaultValue was not registering the change when this.props.defaultValue was cleared to an empty string, which makes sense because why have a default value of nothing?

I'm going to change the LabelTextInput prop name to value instead of defaultValue to make it more clear.

src/Screens/OperatorPanel.tsx
goldjay: Update: I don't think getDerivedStateFromProps is the way to go. To make the decision to clear notes, I need both the current props and the incoming props. Since getDerivedStateFromProps is a static method, I'm unable to access the current props to compare. It seems like getDerivedStateFromProps is better suited for an instance where you have props that will get directly placed into state, since you have access to nextProps and the current state aka the before and after.

auderephilip: Oooh, sorry, I was wrong about this.  I confused `componentDidUpdate` (still good!) with `componentDidReceiveProps`.

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FjUwpNzg9IcyrK%2Fgiphy.gif&token=homer" data-gitmeme-token="homer"></a>

src/Components/LabelTextInput.tsx
auderephilip: Sounds good;  not sure I've thought this through correctly anyway.  If it works for the cases we need it to, no need to change anything.  Thanks!

PR #54, opened 2019-10-21
Found Entry Does Not Appear First

goldjay:
[FEV-1345](https://auderenow.atlassian.net/browse/FEV-1345)

Question: There is still a bit of weirdness in search because of my approach searching all keys and values. EX: If you search a number like 15, you could get an age of 15, a 15 in the timestamp, a 15 in the ID. This makes it pretty difficult searching for things like age or sex. Would it be an improvement to only query specific keys excluding those less likely to be searched?


goldjay:
@balderdash 
As it is, it's possible for two different users to see a different ordering of entries, but the entries stay the same.

Do you mean search results would be displayed separately from the entries so that entries appear twice or that entries would be divided into two panels: one for those that meet the search criteria and ones that do not?

tpaik18:
@goldjay Instead of moving the matching entry to the top, would another possible solution be to open the "Show 8 more" area so it shows everything and then highlight the matching text?

balderdash:
Er sorry, I meant specifically the entries that get sampled, and thus show up before you hit the "Show More" button can change. There was some concern earlier about keeping those consistent across auditors, but that may be less important now that we have the "Show More" button anyway.

Still, the modification based on search feels weird, I normally expect searching to be a read-only kind of operation. I like @tpaik18's suggestion.

goldjay:
@balderdash @tpaik18 Closing this PR in favor of another one With Terri's suggestion.

PR #56, opened 2019-10-22
Flatten backend data collections

auderephilip:
This implements what we talked about last week, where all tasks now reside in the `tasks` collection, changing state instead of being copied back & forth between different collections.

* Code greatly simplified in corestore, now that it can handle tasks generically.
* Auditor's `entries` editing moved out into the AuditorPanel to maintain corestore genericness
* TaskChangeRecord now not replicated in Task.  This simplifies and normalizes the data model, but required that panels load these change records themselves instead of counting on the replica within each Task.
* CSV upload modified to conform to new data model
* Firebase rules updated to protect reads/writes in the correct way
* Renamed the "operator" state into `FOLLOWUP`,  which I think better reflects what that state actually is.

This change results in no functionality difference -- it changes only the way we store data.  Once this lands, I'll do a separate diff to remove collections no longer used.

Here's how the new `tasks` collection looks after CSV upload:
image

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -3,6 +3,12 @@ service cloud.firestore {
     match /actively_viewed_tasks/{docId} {
       allow read, write: if true
     }
+    match /tasks/{docId} {
+      allow read, write: if request.auth.token.roles.hasAny(['Admin'])
+      allow read, write: if request.auth.token.roles.hasAny(['Auditor']) && (request.resource.data.state == "AUDIT" || resource.data.state == "AUDIT")
+      allow read, write: if request.auth.token.roles.hasAny(['Payor']) && (request.resource.data.state == "PAY" || resource.data.state == "PAY")
+      allow read, write: if request.auth.token.roles.hasAny(['Operator']) && (request.resource.data.state == "FOLLOWUP" || resource.data.state == "FOLLOWUP")
+    }
     match /task_changes/{docId} {
       allow read, write: if true
     }
--- a/firestore.rules
+++ b/firestore.rules
@@ -5,7 +5,7 @@ service cloud.firestore {
     }
     match /tasks/{docId} {
       allow read, write: if request.auth.token.roles.hasAny(['Admin'])
-      allow read, write: if request.auth.token.roles.hasAny(['Auditor']) && (request.resource.data.state == "AUDIT" || resource.data.state == "AUDIT")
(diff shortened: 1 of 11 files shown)


src/sharedtypes.ts
```diff
@@ -6,15 +6,20 @@
...
 export const METADATA_COLLECTION = "metadata";
+export const TASKS_COLLECTION = "tasks";
+
+export enum TaskState {
+  CSV = "CSV",
+  AUDIT = "AUDIT",
+  FOLLOWUP = "FOLLOWUP",
+  PAY = "PAY",
+  COMPLETE = "COMPLETE",
+  REJECT = "REJECT"
```
tpaik18: I feel like COMPLETED and REJECTED would be clearer, since they reflect the action that has been done and not the pending action, like AUDIT, FOLLOWUP, and PAY.

src/store/corestore.ts
```diff
@@ -52,214 +47,37 @@ export async function userRoles(): Promise<UserRole[]> {
...
-}
-
-export async function saveOperatorRejectedTask(task: Task, notes?: string) {
-  const change: TaskChangeMetadata = {
+    state: newState,
+    fromState: task.state,
     timestamp: Date.now(),
     by: getBestUserName(),
-    desc: TaskDecision.TASK_REJECTED,
+    desc: `Changed to ${newState}`,
```
tpaik18: Is this the actual string displayed to users in the task history/notes section?

firestore.rules
```diff
@@ -3,6 +3,12 @@ service cloud.firestore {
     match /actively_viewed_tasks/{docId} {
       allow read, write: if true
     }
+    match /tasks/{docId} {
+      allow read, write: if request.auth.token.roles.hasAny(['Admin'])
+      allow read, write: if request.auth.token.roles.hasAny(['Auditor']) && (request.resource.data.state == "AUDIT" || resource.data.state == "AUDIT")
```
balderdash: Auditors also need access to tasks in the `COMPLETE` or `REJECT` state.

src/sharedtypes.ts
auderephilip: Great suggestion.  I've pushed an update to reflect this.

src/store/corestore.ts
auderephilip: Yes - so I thought being a bit more descriptive would help.

firestore.rules
auderephilip: Thanks for pointing this out!  Pushed an update to rules to reflect this.

PR #58, opened 2019-10-22
Deduplicate uploaded records

auderephilip:
This implements full deduplication of CSV uploads on a row-by-row basis by using meta:instanceID from ODK as the unique identifier of records.

* We simplify and flatten the "uploaded records" structure so that one row for each record ever uploaded is stored in the `uploaded_records` collection.
* We scan that collection for newly uploaded meta:instanceIDs, and avoid reinserting IDs we've seen in the past.
* We regenerate a single admin log for each case where a CSV repeats a previously-seen ID.

Note that the "did I previously upload this" search is linear, leading to an O(n^2) implementation of de-duping when it comes to firestore searches.  This is because firestore does not support the concept of querying for records where any of a set of IDs are looked for.  An alternative, faster implementation would be to store all record IDs in one array, since Firestore has an `array-contains` query capability;  however, firestore arrays are limited to a certain modest size, so we wouldn't want to hit that limit.

Here's what the admin record looks like when you have dupes:
image

Here's proof that we only insert non-dupe IDs.  I uploaded the same CSV, but only changed two records (into store 24 and 25).  The other records aren't reinserted, and the admin record shows the other records as rejected.
image

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -25,11 +25,17 @@ import {
 // team's LastPass ServerInfrastructure section.
 const serviceAccount = require("../flows-app-staging-key.json");
 
-const CSV_UPLOAD_COLLECTION = "csv_uploads";
-const CSV_UPLOAD_RECORDS_COLLECTION = "records";
+const UPLOADED_RECORDS_COLLECTION = "uploaded_records";
 
 const ROW_GROUP_BY_KEY = "g3:B01 Pharmacy name";
 const RECORD_ID_FIELD = "meta:instanceID";
+
+type RecordUploadLog = {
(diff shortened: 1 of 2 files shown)


auderephilip:
@tpaik18 Excellent question. I had originally designed the CSV upload to asynchronously process the CSV (via storage hooks on the server) because it enabled us to upload a CSV directly via the console when the UI wasn't yet available.

Now that we've gotten the UI working, I'll make it synchronous so that we can send an error back to the user.

auderephilip:
#64 addresses the synchronicity now.  Thanks again for suggesting!

PR #63, opened 2019-10-22
TaskList Expands to Fill Screen When No Results Are Given

goldjay:
[FEV-54](https://github.com/AudereNow/flows/pull/54)

Overrode some styling with exact pixel values: a bad move, but since the container is "display: relative" and the calendar is "display:absolute" there is nothing we can do in terms of z-index and having the calendar as part of the flow would push all of the search bars, download CSV button, etc down.

No Results Date Picker

### DIFF
--- a/src/Components/DateRangePickerOverride.css
+++ b/src/Components/DateRangePickerOverride.css
@@ -1,3 +1,12 @@
 .DateRangePickerInput {
   overflow: hidden;
 }
+
+.DateRangePicker {
+  position: static;
+}
+
+.DateRangePicker_picker {
+  left: 30px !important;
+  top: 185px !important;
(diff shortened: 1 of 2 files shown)


tpaik18:
Hi @goldjay. I'm running from source now and I feel like width: 300px might be better. The 20% width seems cramped when I run my browser at its usual non-maximized size. Also currently the Payor tab doesn't look right:
Screen Shot 2019-10-23 at 8 58 20 AM

I tried width:300px locally and it fixes the Payor tab issue.

goldjay:
@tpaik18 Oof, that Payor tab looks pretty bad. I'll change it to a set 300px width and tag you on the PR.

PR #65, opened 2019-10-22
Show relative times

auderephilip:
Instead of dates / hours, show a relative time for the user-visible audit trail.

image


auderephilip:
Great suggestions, @goldjay & @tpaik18 !  I've made both edits.  We now use `moment`, and also do on-hover tooltips:

image

tpaik18:
Oh wow, moment.fromNow() is so cool! It has a rounding algorithm that I feel matches humans' mental rounding more closely.

auderephilip:
I had no idea either.  Thanks, @goldjay!

PR #68, opened 2019-10-23
Show full screen spinner when processing CSV

auderephilip:
image

I originally tried to limit the change only to TopBar, using `position:fixed` in hopes of overlaying the whole screen, but the spinner got limited to just the menu itself.  Does anyone know whether, perhaps, something like `float` would make it possible to make the spinner's div paint over the whole screen?  That'd be ideal from an encapsulation perspective.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "react-dates": "^21.2.1",
     "react-dom": "^16.9.0",
     "react-firebaseui": "^4.0.0",
+    "react-loading-overlay": "^1.0.1",
     "react-pose": "^4.0.9",
     "react-scripts": "3.1.1",
     "react-tabs": "^3.0.0",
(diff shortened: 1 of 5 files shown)


auderephilip:
> ... it looks like the loading overlay is already covering the entire screen with your left: 0, etc.

Right now, the PR'd implementation does indeed work correctly.  But in my ideal world, I'd get this effect even if the Loading component were only to appear as a child of TopBar... which I think might still be possible with some form of `float` or some such.  I'll dork around with it a little bit to see if it's possible to get that to work...

auderephilip:
Ok, actually legitimately fixed this now so that encapsulation is preserved.  No changes to MainChrome, no nasty message passing.  Just sweet, sweet spinning.

PR #73, opened 2019-10-24
Highlight relevant text and show all entries on search term found

goldjay:
[FEV-1345](https://auderenow.atlassian.net/browse/FEV-1345)

Behavior:
* Case-sensitive search
* Filters to search based on specific fields

Screen Shot 2019-10-29 at 10 33 10 AM

### DIFF
--- a/src/Components/HighlightText.tsx
+++ b/src/Components/HighlightText.tsx
@@ -1,0 +1,49 @@
+import React from "react";
+import "./HighlightText.css";
+
+function highlight(text: string, searchTerm: string) {
+  let res = [];
+
+  let right = 0;
+  let left = 0;
+  while (right <= text.length - searchTerm.length) {
+    const currTerm = text.substring(right, right + searchTerm.length);
+    if (currTerm === searchTerm) {
+      if (left !== right) {
+        res.push(text.substring(left, right));
+      }
+      res.push(
+        <span key={left} className="highlight">
+          {searchTerm}
+        </span>
+      );
+      right += searchTerm.length;
+      left = right;
+    }
+    right += 1;
+  }
+
+  if (left < text.length) {
+    res.push(text.substring(left));
+  }
+
+  return res;
+}
+
+interface Props {
+  text: string;
+  searchTerm?: string;
+  className?: string;
+}
+
+const HighlightText = (props: Props) => {
+  return (
+    <div className={props.className}>
+      {!!props.searchTerm
+        ? highlight(props.text, props.searchTerm)
+        : props.text}
+    </div>
+  );
+};
+
+export default HighlightText;
--- a/src/Components/HighlightText.tsx
+++ b/src/Components/HighlightText.tsx
@@ -4,29 +4,22 @@ import "./HighlightText.css";
 function highlight(text: string, searchTerm: string) {
   let res = [];
 
-  let right = 0;
-  let left = 0;
-  while (right <= text.length - searchTerm.length) {
-    const currTerm = text.substring(right, right + searchTerm.length);
-    if (currTerm === searchTerm) {
-      if (left !== right) {
-        res.push(text.substring(left, right));
-      }
+  let divided = text.split(searchTerm);
+
+  for (let i = 0; i < divided.length; i++) {
+    const piece = divided[i];
+    if (piece.length > 0) {
+      res.push(piece);
+    }
+
+    if (i < divided.length - 1) {
       res.push(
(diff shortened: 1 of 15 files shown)


src/Components/HighlightText.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import "./HighlightText.css";
+
+function highlight(text: string, searchTerm: string) {
```
audererob: I think you could do text.split(searchTerm), pushing each result, and in between each one insert the highlight span.

src/Screens/AuditorPanel.tsx
```diff
@@ -169,16 +174,22 @@ export class AuditorDetails extends React.Component<Props, State> {
   };
 
   render() {
-    const { showAllEntries } = this.state;
-    const { task, changes } = this.props;
+    const showAllEntries =
+      !!this.props.searchTermGlobal || this.state.showAllEntries;
```
audererob: I think this will result in not being able to collapse the expanded entries despite the UX indicating such. Maybe we also disable or hide that button?

src/Components/HighlightText.tsx
goldjay: There's some trickiness when the search term falls at the beginning or end of the text. I'll mess around with it a bit more since it would be much cleaner.

src/Screens/AuditorPanel.tsx
goldjay: That's a good point. I'll hide the button.

src/Components/HighlightText.tsx
audererob: It seems that an empty string gets returned preceding or following the search term:
"search".split("search") returns ["", ""]

src/Screens/AuditorPanel.tsx
```diff
@@ -169,24 +174,30 @@ export class AuditorDetails extends React.Component<Props, State> {
...
+
           <input
             type="text"
             onChange={this._handleSearchTermDetailsChange}
             placeholder="Filter Details"
           />
         </div>
         {samples.map(this._renderClaimEntryDetails)}
-        {remaining > 0 && (
+        {remaining > 0 && !showAllEntries && (
```
audererob: Assuming we want to keep collapse functionality, should this instead be {remaining > 0 && !searchTermGlobal && (... ?

goldjay: Ah, okay. It works with remaining > 0 && showAllEntries

src/Components/TextItem.css
```diff
@@ -1,8 +1,12 @@
...
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  align-items: center;
   margin: 5px;
 }
 
 .textitem_key {
   font-weight: bold;
+  margin-right: 5px;
```
tpaik18: @goldjay Can we make the text white also? I think that would show up better than black.

goldjay: Sure, no problem.

src/Screens/OperatorPanel.tsx
```diff
@@ -5,12 +5,18 @@ import ImageRow from "../Components/ImageRow";
...
+import { Filters } from "./TaskPanel";
 
 type Props = {
   task: Task;
   notesux: ReactNode;
+  filters: Filters;
+  searchTermGlobal?: string;
+  notes: string;
+};
+type State = {
```
audererob: It doesn't look like state is used in this component.

goldjay: Yes, must have gotten left it when I was resolving conflicts. I'll take it out.

src/Screens/OperatorPanel.tsx
```diff
@@ -54,26 +57,56 @@ export class OperatorDetails extends React.Component<Props> {
...
+        <ImageRow
+          searchTermGlobal={searchTermGlobal}
+          filters={filters}
+          images={this._extractImages(entry)}
+        />
+
+        <ImageRow
+          searchTermGlobal={searchTermGlobal}
+          filters={filters}
+          images={this._extractImages(entry)}
```
tpaik18: @goldjay I noticed that now the OperatorPanel shows each row of images twice. Seems like maybe one of the `ImageRow`s should be removed?

goldjay: Oh no! I'll put the change in the PR I am working on now.

PR #74, opened 2019-10-25
Use tab name  + timestamp as filename for download

tpaik18:
Before: 
Downloaded CSV has names like
```
24-nlz9XiPxYegdnQSllwCw.csv
Random 8-0lXBCU3qBHRd0Lhda6Pg.csv
Random 17-00uZLXOS7CbOKfqIEY7i.csv
22-PEkp65DsBHkI8JYkLs3v.csv
```

After: 
Downloaded CSV has names like
```
Operator_20191024_171705.csv
Complete_20191024_170347.csv
Rejected_20191024_170340.csv
Payor_20191024_170335.csv
Auditor_20191024_170331.csv
```

Please LMK if there's a smarter way to lookup the current tab name.


tpaik18:
> Is the hasOwnProperty line necessary given triple ===? Unsure myself; depends if you expect both to be undefined or some such at some point...

Thanks @auderephilip. I thought that if the Admin tab moved to being the first tab, then the cast to TaskConfig might blow up if I didn't confirm that the tab has the taskState property. But I just tested moving Admin tab to the first and you're right that hasOwnProperty is unnecessary.

PR #77, opened 2019-10-29
One ListItem class instead of three Item classes

tpaik18:
AuditorItem and OperatorItem had identical code. PayorItem was the same except didn't have "5 Claims" in the upper right. 
Removing the three Item classes in favor of one ListItem class. 
Also fix "1 Claims"


src/Components/ListItem.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { Task } from "../sharedtypes";
+import { formatCurrency } from "../store/corestore";
```
audererob: Seems like for safety you should import "../Screens/MainView.css"

tpaik18: @audererob The styles that ListItem needs from MainView.css are only needed by ListItem. So I extracted these into a ListItem.css which seems like better style. Okay?

src/Screens/MainView.tsx
```diff
@@ -26,9 +27,9 @@ const PanelComponents: {
 const ItemComponents: {
```
balderdash: This whole thing is a little bit unnecessary now, but we can keep it around just in case we decide to add a new type of list entry in the future.

At the moment though, `ItemComponents` can just be a an object with a single entry: 
```
{
  default: ListItem
}
```

And in the config, all five tab configs can just say "default" for their `taskListComponent`, rather than specifying "AuditTask", "PayorTask", etc.
https://github.com/AudereNow/flows/blob/5ef3d0e83325853dcaa8b330b85462f427d01eed/src/store/config.ts#L33

src/Components/ListItem.tsx
audererob: Great!

PR #78, opened 2019-10-29
Make buttons configurable

balderdash:
This PR adds an `actions` section to the config that specifies button labels and target task states. The TaskPanel then creates passes the appropriate buttons through to the details component, which renders them as children.

Auditor and Payors do a little extra work when a task gets approved (marking certain entries as reviewed, and actually sending money, respectively), so they register a callback with TaskPanel that runs before the task gets moved to the next state. In the Auditor case, the callback even modifies the task object.


src/store/config.ts
```diff
@@ -33,36 +38,66 @@ export const defaultConfig: AppConfig = {
...
     Payor: {
       taskState: TaskState.PAY,
       taskListComponent: "PayorTask",
       detailsComponent: "PayorTask",
       listLabel: "ITEMS TO REVIEW",
-      roles: [UserRole.PAYOR]
+      roles: [UserRole.PAYOR],
+      actions: {
+        decline: {
+          label: "Decline Payent",
```
tpaik18: "Payment"

src/store/config.ts
```diff
@@ -33,36 +38,66 @@ export const defaultConfig: AppConfig = {
...
+          nextTaskState: TaskState.FOLLOWUP
+        },
+        approve: {
+          label: "Issue Payment",
```
tpaik18: This seems to drop the previous feature of showing the button label as "Mark Paid" if realPayments == false and "Issue Payment" if realPayments==true. Intentional?

src/Screens/OperatorPanel.tsx
```diff
@@ -1,32 +1,19 @@
...
 import { ClaimEntry, Task, TaskState } from "../sharedtypes";
 import { formatCurrency, changeTaskState } from "../store/corestore";
+import { DetailsComponentProps } from "./TaskPanel";
 import "./MainView.css";
 
-type Props = {
-  task: Task;
-  notesux: ReactNode;
+type State = {
   notes: string;
```
audererob: It doesn't look like state is getting used here.

src/Screens/TaskPanel.tsx
```diff
@@ -8,23 +8,34 @@ import LabelWrapper from "../Components/LabelWrapper";
...
+import { ActionConfig } from "../store/config";
 import debounce from "../util/debounce";
 import { containsSearchTerm, DateRange, withinDateRange } from "../util/search";
 import { TaskConfig, defaultConfig } from "../store/config";
 import "./MainView.css";
 
+export interface DetailsComponentProps {
+  task: Task;
+  notesux: ReactNode;
+  notes: string;
```
audererob: I don't think this is needed any more - only DetailsWrapper seems to use it now.

src/store/config.ts
balderdash: It was intentional, but I just added a way to hide/show buttons based on a live config value in order to bring back that functionality.

I'm not sure if it's overkill though, @audererob what do you think about https://github.com/AudereNow/flows/pull/78/commits/09ba63a792933c3dbb55c00a3ac0740864acaf0f?

audererob: I like it - we could alternately have a callback to get the button label, or specify it in the call to registerActionCallback, e.g. genericize it to be something like registerActionProps("approve", { onAction: this._issuePayment, label: realPayment ? "Issue Payment" : "Mark Paid" } );

balderdash: Yeah, I was weighing that kind of approach as well. It would certainly be more flexible, I was just worried about spreading the config across multiple places.

I'll go with this for now, and we'll see how it work out.

PR #79, opened 2019-10-29
FEV-1360 Pop confirmation when switching tabs and entering search UX

audererob:
[FEV-1360](https://auderenow.atlassian.net/browse/FEV-1360)
Follow-up fix for two cases and one bug:
* Case 1 - tab switching wouldn't pop the confirmation. It now does.
* Case 2 - using search/filter UX wouldn't pop the confirmation. It now does.
* Bug - performing an action on the task item that removed it from the list kept the notes text around. It's now cleared.


src/Screens/TaskPanel.tsx
```diff
@@ -120,7 +135,9 @@ export default class TaskPanel extends React.Component<Props, State> {
   };
 
   _onSearchClick = () => {
-    this.setState({ showSearch: !this.state.showSearch });
+    if (this.state.showSearch || this._okToSwitchAway()) {
```
tpaik18: Is there a reason we need the `this.state.showSearch ||`? 
If we just have `this._okToSwitchAway()` then we can avoid losing the note when the user closes the search box.

audererob: Good catch - I added that before fixing the notes clearing problem.

src/Screens/MainView.tsx
```diff
@@ -49,11 +49,25 @@ class MainView extends React.Component<Props, State> {
...
+    const result = !this._onTabSelectCallback || this._onTabSelectCallback();
+    if (result) {
+      this._onTabSelectCallback = undefined;
+    }
+    return result;
+  };
+
+  _registerForTabSelectCallback = (onTabSelect: () => boolean) => {
+    this._onTabSelectCallback = onTabSelect;
+  };
```
auderephilip: I found the semantics of this diff a bit tricky to understand, but I think now I get it.  I'm not sure it could have been done better, so perhaps there's just some basal complexity that can't be gotten rid of... but here are some small thoughts / suggestions / comments:

* As a first-time reader to this code (not from this PR, but directly from MainView, it took me a bit of digging to figure out the semantic meaning of "result" (i.e. it ultimately tells react-tabs whether to allow navigation).  I suspect if we named the local variable something like `canSwitchTabs`, I might have caught on earlier.
* The unregistration (line 62) also confused me for a second... because it takes a while to see that it's probably safe to do it there because react-tabs is going to navigate away, and so tabs will need to reregister later.  (Do they do that correctly?  Like if their other state has been preserved, because they haven't been unmounted?).
* The semantics of having each Panel now know that they're hosted in Tabs (because they need to register for tab select events) seems a bit like it breaks some form of encapsulation.  But I'm not sure if it could be done better;  I'm opining here because I'm about to integrate something like `react-router` to make top-level permalinks work in the app, which might then suggest some form of analagous `canNavigateAway` callback each panel would have to support.

I don't actually have a concrete suggestion about how to make this better;  this comment is more meant as a conversation starter so that my integration of `react-router` doesn't make this even more confusing.  ;)

PR #80, opened 2019-10-29
Add lastUpdate field to Task for sorting

tpaik18:
Adds a lastUpdate field to Task so that the list of tasks won't be in random order. 
Not sure if this is the right approach to sorting the list.


src/sharedtypes.ts
```diff
@@ -76,6 +76,7 @@ export type Task = {
   state: TaskState;
   entries: ClaimEntry[];
   site: Site;
+  lastUpdate: number; // timestamp
```
auderephilip: Opinion Q: `lastUpdate` is indeed a much better name than `timestamp` in the other objects right now... but the others are all named `timestamp`.  ;)  Classic debate between consistency and correctness.  I think keeping this is just fine;  no one will be confused.

tpaik18:
@auderephilip Oh, thanks! I'll look into event triggers. That does seem like the Awesomer way.

src/sharedtypes.ts
tpaik18: @auderephilip The reason I didn't use `timestamp` was because I felt the timestamp for a Task could either mean the first or last timestamp of its ClaimEntries *or* the latest timestamp of its associated `task_changes`. At first I tried `orderBy("entries[0].timestamp")` which did not work. There could be an argument for sorting by the entries' timestamp--prompting the user to act on the oldest claims first? But then I thought the email inbox paradigm was better where most recently updated should bubble to the top... Thoughts? 

At any rate, it does seem like `updatedAt` is a more standard name than `lastUpdate`. Welcome other opinions on `updatedAt` vs `timestamp`.

auderephilip: Sounds great to me.  `updatedAt` feels a bit better IMO than `lastUpdate`... in that I could see a `createdAt` in the future, whereas it's not as clear to me what the grammatic parallel to `lastUpdate` would be for creation.

tpaik18:
Okay, I think I finally implemented the trigger approach. 
The internet recommended `firebase.firestore.FieldValue.serverTimestamp()` for the timestamp, but that's an object and not a number, and it kept failing the `===` comparison. So I just used Date.now() so that I'd get a simple number field that I could check `===` against. (The equality check is needed to prevent an infinite loop of updating the `updatedAt` field causing the trigger to run again.) 

The sort by updatedAt is commented out for now, because then records without `updatedAt` will not show up in the UI. If we clear out the database and start fresh, then we can sort by updatedAt.

tpaik18:
In response to @auderephilip's comment about writing triggers for any object, yes, I think we could, but it seems `tasks` are the only objects that get updated. Do we need triggers for anything else? 

One thought is that we could move the creation of the task_change record into the trigger for task update.

functions/src/index.ts
```diff
@@ -174,6 +174,24 @@ exports.uploadCSV = functions.https.onCall(
   }
 );
 
+exports.onUpdateTask = functions.firestore
+  .document(TASKS_COLLECTION + "/{task}")
+  .onWrite((change, context) => {
+    if (
+      change.after.exists &&
+      (!change.before.exists ||
+        change.after.data()!.updatedAt === change.before.data()!.updatedAt)
```
auderephilip: Oh, interesting.  So this allows the client to override this behavior by insisting on a specific updatedAt?  Wanted to confirm that's what you want.

functions/src/index.ts
```diff
@@ -174,6 +174,24 @@ exports.uploadCSV = functions.https.onCall(
...
+    ) {
+      const timestamp = Date.now();
+      return admin
+        .firestore()
+        .collection(TASKS_COLLECTION)
+        .doc(context.params.task)
```
auderephilip: tiny thought for the future:  perhaps `taskID` as a name when it's an ID, vs a full task, is at times useful. NBD here -- just a thought for the future.

auderephilip: Oooh - one other thought:  it looks from https://firebase.google.com/docs/functions/firestore-events that we should instead do `change.after.ref.set({...})` here.

functions/src/index.ts
tpaik18: @auderephilip the after.updateAt === before.updateAt check is to prevent infinite loop. The update of the timestamp is also a write to the document, so if this trigger is firing because of the update of the timestamp, then do not update the timestamp again.

auderephilip: Ah, righto.  Sorry, should have read your description more carefully... you had already covered that.

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FjUwpNzg9IcyrK%2Fgiphy.gif&token=homer" data-gitmeme-token="homer"></a>

functions/src/index.ts
tpaik18: Oh nice! I verified that it works. Thanks.

PR #84, opened 2019-10-30
Search Bar Appears at Top of TaskList when scrolling

goldjay:
[FEV-1374](https://auderenow.atlassian.net/browse/FEV-1374)

TIL about position: sticky

### DIFF
--- a/src/Screens/TaskPanel.tsx
+++ b/src/Screens/TaskPanel.tsx
@@ -421,6 +421,8 @@ export default class TaskPanel extends React.Component<Props, State> {
         />
       ) : null;
 
+    const renderSearchFn = !!showSearch ? this._renderSearchPanel : undefined;
+
     return (
       <div className="mainview_content">
         <LabelWrapper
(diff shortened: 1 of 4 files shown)


src/Screens/TaskPanel.tsx
```diff
@@ -421,15 +421,17 @@ export default class TaskPanel extends React.Component<Props, State> {
...
     return (
       <div className="mainview_content">
         <LabelWrapper
           label={`${this.props.listLabel}: ${this.state.tasks.length}`}
           className="mainview_tasklist"
           renderLabelItems={this._renderLabelItems}
         >
-          {!!showSearch && <div>{this._renderSearchPanel()}</div>}
           <TaskList
+            renderSearchPanel={renderSearchFn}
```
auderephilip: Is this a react-standard sort of way to render sub-parts of a component, or do people instead pass `JSX.Element` or Components as a prop?  (I have no idea - just asking).

goldjay: You know, I'm not sure what is the react-standard way. I've seen people using this.props.children when there are many different sub-parts subject to change and passing a component when you would need the sub-part to handle its own re-render independent of its parent. I'll look into it a bit before merging this PR.

PR #86, opened 2019-10-30
Case-Insensitive Search

goldjay:
[FEV-1369](https://auderenow.atlassian.net/browse/FEV-1369)

### DIFF
--- a/src/Screens/TaskPanel.tsx
+++ b/src/Screens/TaskPanel.tsx
@@ -9,11 +9,7 @@ import Notes from "../Components/Notes";
 import TaskList from "../Components/TaskList";
 import { Task, TaskChangeRecord, TaskState } from "../sharedtypes";
 import { ActionConfig, defaultConfig, TaskConfig } from "../store/config";
-import {
-  changeTaskState,
-  getChanges,
-  subscribeToTasks
-} from "../store/corestore";
+import { changeTaskState, getChanges, subscribeToTasks } from "../store/corestore";
 import { getConfig } from "../store/remoteconfig";
 import debounce from "../util/debounce";
 import { containsSearchTerm, DateRange, withinDateRange } from "../util/search";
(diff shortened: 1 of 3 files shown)


src/Screens/TaskPanel.tsx
```diff
@@ -262,7 +262,7 @@ export default class TaskPanel extends React.Component<Props, State> {
     this.setState({
       searchDates: { startDate: null, endDate: null },
       tasks: allTasks,
-      filters: { patient: true, name: true, patientID: true, item: true },
+      filters: { patient: false, name: false, patientID: false, item: false },
```
auderephilip: Didn't we just make this change somewhere?  Looks really familiar.

src/Screens/TaskPanel.tsx
```diff
@@ -421,6 +421,10 @@ export default class TaskPanel extends React.Component<Props, State> {
       ) : null;
 
     const searchPanel = !!showSearch ? this._renderSearchPanel() : undefined;
+    // let filters = this.state.filters;
+    // if (!Object.values(filters).some(value => !!value)) {
+    //   filters = { patient: true, name: true, patientID: true, item: true };
+    // }
```
auderephilip: Not sure I fully understand this.  Thought the other PR wanted this.

src/util/search.ts
```diff
@@ -26,11 +26,10 @@ export const containsSearchTerm = (
   for (let i = 0; i < filterKeys.length; i++) {
     const filterValue = (filters as any)[filterKeys[i]];
     const filterKey = filterKeys[i];
-
     if (
       !!filterValue &&
       !!entryCopy[filterKey] &&
-      entryCopy[filterKey].includes(searchPhrase)
+      entryCopy[filterKey].toLowerCase().includes(searchPhrase.toLowerCase())
```
auderephilip: Recommend lowercasing searchPhrase once outside this `for` loop.

src/Screens/TaskPanel.tsx
goldjay: I had a previous PR where I made the default as true, but since I changed the no filter behavior to all filters, and most people will only want to search by 1 or 2 filters, I thought the default state should be false.

src/Screens/TaskPanel.tsx
goldjay: Just some leftovers from git weirdness.

PR #88, opened 2019-10-31
Update app name/title

auderephilip:
A bit funny we were still "React App".  :D

image


auderephilip:
Great idea on favicon, @tpaik18 . I executive-decisioned it and instead put Audere's logo in the favicon;  been talking with David about balancing branding, between the experience being Maisha and a subtle small brand of Audere in the background so it's clear we're hosting the service.  We can always change this later.

PR #89, opened 2019-10-31
Show admin full history

auderephilip:
Yes, it's nasty, but we'll soon design this to actually be usable (e.g. look good, present notes only when relevant, permalink to tasks for drill-down, show dates readably, be sortable / searchable, pagination).  For now, this gets the data in there and lets us enhance it in parallel.

image

### DIFF
--- a/src/Screens/AdminPanel.tsx
+++ b/src/Screens/AdminPanel.tsx
@@ -2,6 +2,7 @@ import firebase from "firebase";
 import React from "react";
 import { UserRole } from "../sharedtypes";
 import { setRoles } from "../store/corestore";
+import ChangeHistory from "../Components/ChangeHistory";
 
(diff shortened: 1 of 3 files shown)


src/Screens/AdminPanel.tsx
```diff
@@ -27,8 +28,8 @@ class AdminPanel extends React.Component<Props, State> {
   };
 
   async componentDidMount() {
-    const adminLogs = await this._getAdminLogs();
-    console.log(adminLogs);
+    /*    const adminLogs = await this._getAdminLogs();
+    console.log(adminLogs);*/
```
auderephilip: (this commented out because @goldjay has some pending work on this... leaving it uncommented actually causes an error on the admin page)

PR #90, opened 2019-10-31
Firestore fixes for admin log events in client

auderephilip:
This enables us to work on AdminLogEvents in the client a little more:
* We shouldn't include firebase directly, but instead should import frome firebase/app and also specific firebase modules (like firebase/firestore)
* We should put all firestore access in `corestore` so that we can easily refactor storage in an app-independent way.
* This also updates firestore rules so that the admin panel can actually load history

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -12,6 +12,9 @@ service cloud.firestore {
     match /task_changes/{docId} {
       allow read, write: if true
     }
+    match /admin_log_event/{docId} {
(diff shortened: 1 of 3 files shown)


auderephilip:
@goldjay : I know you're mid-action on some of this stuff;  hopefully this helps move it along a bit!  I was just in the admin panel today, so thought I'd send this along.

PR #91, opened 2019-10-31
Show all people working on task

tpaik18:
Before this fix, the tooltip did not display all people working on the task. 
Now it does. And it only adds the word "also" if I'm working on the task too. (Nitpicky, I know.)

Screen Shot 2019-10-31 at 10 58 34 AM


src/Components/TaskList.tsx
```diff
@@ -100,11 +101,15 @@ class TaskList extends React.Component<Props, State> {
           <div className="tasklist_search_panel">{this.props.searchPanel}</div>
         )}
         {tasks.map((task, index) => {
-          const activeTask = this._isActiveTask(task);
-          const activeClass = activeTask ? "tasklist_active" : undefined;
-          const activeDataTip = activeTask
-            ? `${activeTask!.name} also working on this task`
-            : undefined;
+          let activeClass = undefined;
```
goldjay: Small nit: variables are undefined by default so it could be: 

let activeClass, activeDataTip;

goldjay:
A bit late on this, but as we get more users it may be nice to limit the number of "currently working on" names in the tooltip to 5 or so.

PR #92, opened 2019-10-31
Enables permalinks to tasks

auderephilip:
This PR is step 1 of allowing url links directly to individual tasks.  Subsequent PRs will change the window url when tasks are selected and also hyperlink tasks.

This uses react-router-dom, and avoids getDerivedStateFromProps using React's recommendation of key-based component updates if initial props need to affect rendering only when those props change.  But I'm open to other approaches if they're cleaner!

Here's what it looks like linking directly to an operator task:
image


src/Screens/TaskPanel.tsx
```diff
@@ -72,8 +77,10 @@ export default class TaskPanel extends React.Component<Props, State> {
   };
   _unsubscribe = () => {};
   _inputRef: React.RefObject<HTMLInputElement> = React.createRef();
+  _useInitialTaskID: boolean = false;
 
   async componentDidMount() {
+    this._useInitialTaskID = !!this.props.initialSelectedTaskID && true;
```
goldjay: Why is the "&& true" here?

auderephilip: Wow - no idea.  Good catch. Removed in subsequent diff.

PR #93, opened 2019-10-31
overflow of date range picker now visible

goldjay:
[FEV-1379](https://auderenow.atlassian.net/browse/FEV-1379)
[FEV-1378](https://auderenow.atlassian.net/browse/FEV-1378)

Removed all of that positioning nonsense and just put the panel in the LabelWrapper with a css hide property that covers FEV-1378.

Screen Shot 2019-10-31 at 11 41 12 AM


src/Components/DateRangePickerOverride.css
```diff
@@ -7,6 +7,5 @@
 }
 
 .DateRangePicker_picker {
-  left: 30px !important;
-  top: 185px !important;
+  top: 110px !important;
```
auderephilip: I unfortunately don't understand CSS well enough, but have always been told that `!important` has bad smells.  What is it here to override?  And is there a way to do it without `!important`?

src/Components/TaskList.tsx
```diff
@@ -99,26 +99,28 @@ class TaskList extends React.Component<Props, State> {
         {!!searchPanel && (
           <div className="tasklist_search_panel">{this.props.searchPanel}</div>
         )}
-        {tasks.map((task, index) => {
-          const activeTask = this._isActiveTask(task);
-          const activeClass = activeTask ? "tasklist_active" : undefined;
-          const activeDataTip = activeTask
-            ? `${activeTask!.name} also working on this task`
-            : undefined;
+        <div style={{ marginTop: !!searchPanel ? "124px" : "0px" }}>
```
auderephilip: This feels like we're mixing inline styles with classnames.  Could we instead make this just use classnames (e.g. `<div className={!!searchPanel ? "withSearch" : "withoutSearch"}>`)

goldjay: Sure, no problem. I'll just make another className and add it to the existing class string conditionally.

src/Components/DateRangePickerOverride.css
goldjay: Normally, yes, !important should be avoid like poison. However, the styling on DateRangePicker_picker is inline, which can only be overridden by !important or javascript replacement shenanigans. Double however: I've changed the parent styling for other tickets in such a way that it isn't needed anymore. Thanks for bringing it up!

PR #94, opened 2019-10-31
Make the favicon show larger in the browser

auderephilip:
Cropped the image so that it fills the extents of the total image dimensions

image


auderephilip:
@tpaik18 : le sigh. I made a file-save error.  Have corrected it now!

PR #95, opened 2019-11-01
Much better TaskChangeRecord rendering

auderephilip:
We integrate `react-table`, which supports column widths, paging, sorting, and a bunch of other things automatically.

image


auderephilip:
Oh - and we might consider whether or not just replacing existing `DataTable` usage with `ReactTable` might be wholesale good.

auderephilip:
@tpaik18 : great idea.  Fixed!

PR #97, opened 2019-11-04
Pass remote config values as props

balderdash:
This adds a `configureComponent()` method that is basically the equivalent of `redux-react`'s `connect()` method, but it ties a component to remote config values rather than redux state. It takes in a component that accepts one or more props that depend on remote config values, a method to map remote config state to those props, and returns a component that only requires the remaining props.

I was able to remove the code in the Payor Details component that explicitly fetched the "realPayments" remote config value, and instead it just accepts it as a prop. The TaskPanel changes were a bit less clean, since the values it needs are based on the config, but it also works there.

I started this PR to address @auderephilip's comment [here](https://github.com/AudereNow/flows/commit/09ba63a792933c3dbb55c00a3ac0740864acaf0f#r35739927), and got a little carried away.

### DIFF
--- a/src/store/remoteconfig.ts
+++ b/src/store/remoteconfig.ts
@@ -39,3 +39,13 @@ export async function getConfig(key: string) {
   }
   console.error(`Didn't find key ${key} in remoteConfig!`);
 }
+
+export function subscribeToConfigs(
+  callback: (config: RemoteConfig) => void
+): () => void {
+  return firebase
+    .firestore()
+    .collection(METADATA_COLLECTION)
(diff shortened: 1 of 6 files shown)


src/store/remoteconfig.ts
```diff
@@ -39,3 +39,13 @@ export async function getConfig(key: string) {
   }
   console.error(`Didn't find key ${key} in remoteConfig!`);
 }
+
+export function subscribeToConfigs(
```
auderephilip: So good.  I never thought of this.

PR #100, opened 2019-11-05
Better error propagation to user

auderephilip:
Two things:
1. When the CSV has issues (e.g. firestore throws), turn the error into something communicated back to the user
2. Catch a bunch of client-side errors that can happen when the server throws.

Addresses user-reported error where CSV processing is silently failing on the server.

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -149,28 +149,35 @@ exports.uploadCSV = functions.https.onCall(
     await LogAdminEvent(user, `Processing CSV upload from ${user.name}`);
     const cache: any[] = [];
 
-    const result: CallResult = await new Promise((res, rej) =>
-      csvtojson({ needEmitAll: true })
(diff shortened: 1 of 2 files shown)


functions/src/index.ts
```diff
@@ -149,28 +149,35 @@ exports.uploadCSV = functions.https.onCall(
...
-    );
-
-    return result;
+          )
+      );
+      return result;
+    } catch (e) {
+      return {
+        error: `CSV processing error: ${e.message ||
+          e.error ||
```
tpaik18: Interesting. So `e.err` and `e.error` both work?

auderephilip: No, unfortunately it depends who's throwing the error.  CSV2JSON throws `e.err`, Firestore seems to throw `e.error`, and many other people throw `e.message`.

PR #108, opened 2019-11-08
Searchable Admin Logs + Change History

goldjay:
[FEV-1399](https://auderenow.atlassian.net/browse/FEV-1399)

* On Change History, not all changes have a fromState and toState (rejected, declined, uploads) so I added some logic to get the user doing the action and the action's description. Other than that, the SearchableTable should be the same as ChangeHistory. 

Screen Shot 2019-11-08 at 10 33 34 AM

### DIFF
--- a/src/Components/SearchableTable.tsx
+++ b/src/Components/SearchableTable.tsx
@@ -1,27 +1,13 @@
+import { json2csv } from "json-2-csv";
 import moment from "moment";
 import React from "react";
-import ReactTable, { RowRenderProps } from "react-table";
+import ReactTable from "react-table";
 import "react-table/react-table.css";
-import ReactTooltip from "react-tooltip";
-import { TaskChangeRecord } from "../sharedtypes";
-import { getAllChanges } from "../store/corestore";
 import debounce from "../util/debounce";
 import { containsSearchTerm } from "../util/search";
 import Button from "./Button";
-import "./ChangeHistory.css";
 import CheckBox from "./CheckBox";
-
-const TABLE_COLUMNS = [
-  { Header: "Task ID", accessor: "taskID", minWidth: 150 },
-  {
-    Header: "Time",
-    accessor: "timestamp",
-    Cell: (props: RowRenderProps) => renderTooltippedTime(props.value),
-    minWidth: 100
-  },
-  { Header: "Description", accessor: "description", minWidth: 450 },
-  { Header: "Notes", accessor: "notes", minWidth: 200 }
-];
+import "./SearchableTable.css";
 
 type ChangeRow = {
   taskID: string;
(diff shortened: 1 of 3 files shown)


src/Components/SearchableTable.tsx
```diff
@@ -30,46 +16,40 @@ type ChangeRow = {
   notes?: string;
 };
 
-type Props = {};
+type Props = {
+  tableColumns: any[];
+  adapterFunction: (all: any[]) => any[];
+  dataFetchingFunction: () => Promise<any[]>;
```
auderephilip: If we're going to turn this into a generic component, it feels cleaner to just make the parent pass pre-fetched `allData`, especially since `dataFetchingFunction` is effectively only use during `componentDidMount`.

src/Components/SearchableTable.tsx
```diff
@@ -98,17 +78,47 @@ class ChangeHistory extends React.Component<Props, State> {
...
 
   _handleSearchChange = debounce((searchTerm: string) => {
-    const filteredChanges = this._computeFilteredChanges(searchTerm);
+    const filteredData = this._computeFilteredChanges(searchTerm);
 
-    this.setState({ changes: filteredChanges, searchTerm });
+    this.setState({ data: filteredData, searchTerm });
   }, 500);
 
+  _downloadCSV = () => {
```
auderephilip: This feels a bit funky to me, that a SearchableTable contains such specific functionality.  But I guess it's not clear how a Download button outside this component would easily be able to get to the filtered data... unless we factored this differently, such that the parent owns both a Table, a search box, and a Download button;  the parent maintains `allData` and `data`, passing only `data` to a generic DataTable component (thus eliminating the concept of search from the table).  But if I guess if we did it that way, it'd be trickier to highlight the search terms within the table.

@balderdash : any opinions on factoring?  It feels weird to me that a searchable table  would contain CSV downloading functionality;  feels better that the data mgmt is hoisted to the parent.  But LMK.

balderdash: It feels funky to me as well, but I can't think of a way to cleanly separate things out.

We could do it the way you suggested, and just have the `Table` component also accept the search term, partially redoing the search in order to show the search term highlights. This is probably the most react-y factoring, but redoing the search is neither particularly clean nor efficient.

Another option would be to replace this method with an `exportFn` prop on `SearchableTable`, so that this component just passes the filtered data to a callback. The csv-specific functionality could then live directly in the Admin Panel component or an intermediate wrapper. This at least separates out the functionality that's most out of place in this component.

auderephilip: Ah, neat idea, @balderdash .  Thanks for the simplification;  perhaps on `onExport` or some such would be a way to make this ultimately cleaner.

It feels like in the meantime we should just leave this alone;  given all the tradeoffs (and none of them both fix the problem truly and are cheap), I'd say we just ship with CSV embedded for now.

PR #109, opened 2019-11-08
Recalculate Changes on Search

goldjay:
[FEV-1412](https://auderenow.atlassian.net/browse/FEV-1412)

### DIFF
--- a/src/Screens/TaskPanel.tsx
+++ b/src/Screens/TaskPanel.tsx
@@ -237,7 +237,7 @@ class TaskPanel extends React.Component<Props, State> {
     });
   };
 
-  _handleSearchTermGlobalChange = debounce((searchTerm: string) => {
+  _handleSearchTermGlobalChange = debounce(async (searchTerm: string) => {
     const { selectedTaskIndex, tasks } = this.state;
     const selectedId =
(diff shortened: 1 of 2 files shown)


src/Screens/TaskPanel.tsx
```diff
@@ -237,14 +237,17 @@ class TaskPanel extends React.Component<Props, State> {
...
+  _handleSearchTermGlobalChange = debounce(async (searchTerm: string) => {
     const { selectedTaskIndex, tasks } = this.state;
     const selectedId =
       selectedTaskIndex >= 0 ? tasks[selectedTaskIndex].id : "";
     const filteredTasks = this._computeFilteredTasks(
       searchTerm,
       this.state.searchDates
     );
+
+    const changes = await Promise.all(filteredTasks.map(t => getChanges(t.id)));
```
auderephilip: Wow.  The implementation of `getChanges` when used like this will become super-inefficient in large lists of tasks.

At minimum, we should probably change the `getChanges` function to take an array of TaskIds, break them into chunks of ten, and use Firestore's `in` operator when querying.  Doesn't need to be done in this PR, though;  I'll file a task.

PR #113, opened 2019-11-19
Merge change history admin logs

goldjay:
[FEV-1465](https://auderenow.atlassian.net/browse/FEV-1465)
[FEV-1433](https://auderenow.atlassian.net/browse/FEV-1433)

* Allow notes text to wrap around instead of being cut off
NOTE: I could also do this for description, but the description for rejected CSVs is very long so it's not practical
* Merged the two admin tables together

Screen Shot 2019-11-19 at 12 34 39 PM


balderdash:
I'm not sure the User ID column is very useful here. Can we just drop it?

goldjay:
@balderdash Sure

PR #115, opened 2019-11-19
FEV-1313 Add pharmacy docs in firestore and editable pharmacy notes

balderdash:
Adds a new collection of pharmacy docs under the `pharmacies` collection in firestore. First use for this is an editable notes field for a pharmacy that persists across tasks.

image

image

### DIFF
--- a/src/Components/PharmacyInfo.tsx
+++ b/src/Components/PharmacyInfo.tsx
@@ -98,7 +98,7 @@ class PharmacyInfoHelper extends React.Component<Props, State> {
             </div>
           ) : (
             <div className="pharmacy_detail">
-              Notes: {this.state.pharmacy.notes}{" "}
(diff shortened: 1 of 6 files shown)


src/Components/PharmacyInfo.tsx
```diff
@@ -0,0 +1,108 @@
...
+                  onClick={this._onNotesSave}
+                  disabled={this.state.saving}
+                >
+                  Save
+                </button>
+              </div>
+            </div>
+          ) : (
+            <div className="pharmacy_detail">
+              Notes: {this.state.pharmacy.notes}{" "}
```
goldjay: Nit: It's a bit strange to have text out in the open and that space hanging out there. Maybe you could do something like <span>{`Notes: ${this.state.pharmacy.notes }`}</span>

PR #116, opened 2019-11-19
FEV-1461 Filter tasks by pharmacy owners

balderdash:
In the "secondary followup" tab, only show ops team members tasks for pharmacies that they own.

Note: This doesn't apply to primary review. We want to show tasks for unowned pharmacies to every member of the team so that they don't fall through the cracks. Reconciling that with a single list of owners gets tricky, so I'm not going to implement that for now. Afkera said they probably won't need this feature in primary review for a while anyway.

### DIFF
--- a/src/Screens/TaskPanel.tsx
+++ b/src/Screens/TaskPanel.tsx
@@ -235,20 +235,6 @@ class TaskPanel extends React.Component<Props, State> {
     if (this.state.disableOwnersFilter) {
       return true;
     }
-    if (!this.state.pharmacies.hasOwnProperty(siteName)) {
-      getPharmacyDetails(siteName).then(pharmacy =>
-        this.setState(
-          state => ({
-            pharmacies: {
-              ...state.pharmacies,
-              [siteName]: pharmacy
-            }
-          }),
-          this._updateTasks
-        )
-      );
-      return true;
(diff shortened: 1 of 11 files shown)


src/Screens/TaskPanel.tsx
```diff
@@ -220,11 +231,37 @@ class TaskPanel extends React.Component<Props, State> {
     this._handleSearchTermGlobalChange(input);
   };
 
+  _checkOwner = (siteName: string) => {
```
goldjay: Something about this feels a little strange. I think it's because checkOwner ends up re-rendering the task panel each time it finds a pharmacy. Is there a way you could get all of the information and just update state once you are done?

src/Screens/TaskPanel.tsx
```diff
@@ -220,11 +231,37 @@ class TaskPanel extends React.Component<Props, State> {
...
+    if (this.state.disableOwnersFilter) {
+      return true;
+    }
+    if (!this.state.pharmacies.hasOwnProperty(siteName)) {
```
goldjay: Something about this feels strange. It would cut down the renders if we only updated state with the pharmacies once we have them all in a local variable and separated that functionality from checkOwner.

PR #126, opened 2019-11-21
Advanced search

goldjay:
[FEV-1467](https://auderenow.atlassian.net/browse/FEV-1467)

New Search!
* TaskPanel by patient(id, name, gender, phone), item or pharmacy name
* History by id, time description, notes

Can use multiple filter keys + search terms ex: item:e-p, patient:ru
Works without any search keys as well and searches all fields
highlighting on the TaskPanel


src/Screens/TaskPanel.tsx
```diff
@@ -446,33 +407,21 @@ class TaskPanel extends React.Component<Props, State> {
         </div>
         <div className="labelwrapper_row">
           <div className="mainview_search_row">
+            <ToolTipIcon
```
balderdash: Can you make this tooltip wider for readability?
image

PR #130, opened 2019-11-24
FEV-1436 Configurable canned responses

balderdash:


src/Components/CannedNotesEditor.tsx
```diff
@@ -0,0 +1,160 @@
...
+    if (!this.state.notes) {
+      return null;
+    }
+    return (
+      <ul>
+        {this.state.notes.map((note, index) =>
+          this.state.editing[index] === undefined ? (
+            <li key={index}>
+              {note}
+              <Button label="Edit" onClick={() => this._editNote(index)} />
```
goldjay: Perf nit: Instead of having the arrow function inside the onClick handler, it may be worth it to pass a "data-value" prop that gets put into the button html element that can be retrieved from the event i.e. event.currentTarget.getAttribute("data-value");

PR #133, opened 2019-11-26
Notes + Rejection for individual claims

goldjay:
[FEV-1404](https://auderenow.atlassian.net/browse/FEV-1404)
[FEV-1435](https://auderenow.atlassian.net/browse/FEV-1435)

* User can Add + Edit individual notes for a claim
* User can Reject an individual claim and it will not be included in the reimbursement total

MISC: 
* Checkboxes disabled on rejected, completed tabs
* Total reimbursement of pharmacy adjusts in realtime when rejecting claims
* Rejected claims show up in a separate table within the payor tab to prevent confusion
Rejected Claims


src/store/corestore.ts
```diff
@@ -449,3 +454,34 @@ export async function getPharmacyClaims(siteName: string) {
...
+}
+
+export async function setRejectedClaim(
+  task: Task,
+  claimIndex: number,
+  rejected: boolean
+) {
+  task.entries[claimIndex].rejected = rejected;
+
+  console.log("SETTING REJECTED CLAIM");
```
balderdash: Clean up log

src/Components/ClaimNotes.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  _onNotesChange = (event: ChangeEvent<HTMLTextAreaElement>) => {
+    this.setState({ notes: event.currentTarget.value });
+  };
+
+  render() {
+    const { notes, editing } = this.state;
+    return (
+      <div className="claimnotes_row">
+        {!!editing ? (
```
balderdash: Nit: `editing` is already a boolean, no need to `!!` it.

PR #134, opened 2019-11-26
Download Rejection Report

goldjay:
[FEV-1463](https://auderenow.atlassian.net/browse/FEV-1463)

Download list of claims for a pharmacy with rejected status and individual notes


.env
```diff
@@ -0,0 +1 @@
+SKIP_PREFLIGHT_CHECK=true
```
balderdash: Did you intend to add this file?

goldjay: I sure didn't!

PR #136, opened 2019-12-03
Redesign of application

goldjay:
Screen Shot 2019-12-03 at 10 00 20 AM
Screen Shot 2019-12-03 at 10 00 29 AM
Screen Shot 2019-12-03 at 10 00 44 AM
Screen Shot 2019-12-03 at 10 00 53 AM
Screen Shot 2019-12-03 at 10 02 12 AM
Screen Shot 2019-12-03 at 10 02 42 AM


src/Components/Button.tsx
```diff
@@ -35,17 +36,20 @@ class Button extends React.PureComponent<Props> {
...
       <button
-        className={`button_container ${enabledClassName} ${className || ""}`}
+        className={`button_container ${enabledClassName} ${className}`}
         disabled={disabled}
         name={name}
         type="button"
         onClick={this._onClick}
       >
+        {!!labelImg && (
+          <img className="button_image_label" alt={labelImg} src={labelImg} />
```
balderdash: `alt` should probably be `label`, not the same as src.

src/store/config.ts
```diff
@@ -11,6 +13,9 @@ interface CustomPanelConfig extends TabConfig {
 
 export interface ActionConfig {
   label: string;
+  labelClassName?: string;
+  labelImg?: string;
+  postLabelElement?: JSX.Element;
```
balderdash: Looks like this is unused.

goldjay: It's used in the task panel to have the number of items to review colored.

balderdash: That's as a prop to `LabelWrapper`, it doesn't also need to be in `ActionConfig` does it?

goldjay: Gotcha. No, it doesn't need to be in ActionConfig.

PR #145, opened 2019-12-05
Tab navigation fix

goldjay:
Needed a way to differentiate the props from a link navigation vs. tab navigation.

### DIFF
--- a/src/Screens/MainView.tsx
+++ b/src/Screens/MainView.tsx
@@ -65,12 +65,15 @@ class MainView extends React.Component<Props, State> {
     this.setState({ roles, selectedTabIndex });
   }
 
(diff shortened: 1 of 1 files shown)


src/Screens/MainView.tsx
```diff
@@ -65,12 +65,15 @@ class MainView extends React.Component<Props, State> {
     this.setState({ roles, selectedTabIndex });
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(nextProps: Props) {
```
balderdash: It feels like this method is providing functionality that we should be getting out of the box from react-router, and the fact that this is necessary means there's something wrong with the way we're using the library.

Oh well 🤷‍♂

PR #153, opened 2019-12-10
FEV-1543 Fix tab index lookup

balderdash:
Tab indices were being looked up with the assumption that every user sees every tab. We need to filter by role before doing the lookup.


goldjay:
There's a bit of weirdness when clicking links from the admin change history.

balderdash:
I wasn't able to repro the missing tab, but I did fix some links from the admin tab.

goldjay:
@balderdash Looks like the tab was an issue with setting my permissions 😅

PR #154, opened 2019-12-10
Production credentials + Scheduled backup

goldjay:
* firebase credentials based on process environment variables
* Scheduled backup + old backup culling functions

NOTE: I'm going to change the backup to weekly once I verify the scheduler works

NOTE 2: Any recommendations for storing the API info other than an external non-check-in file?


functions/src/index.ts
```diff
@@ -26,14 +26,21 @@ import {
 
 // You're going to need this file on your local machine.  It's stored in our
 // team's LastPass ServerInfrastructure section.
-const serviceAccount = require("../flows-app-staging-key.json");
+const serviceAccount =
+  !process.env.NODE_ENV || process.env.NODE_ENV === "development"
```
balderdash: Does this work correctly for staging?

functions/src/index.ts
```diff
@@ -57,6 +64,58 @@ type CallResult =
       result: string;
     };
 
+const bucketName = "gs://flows-app-production-backup";
```
balderdash: I'd prefer not to hardcode "production". How about `gs://${serviceAccount.project_id}-backup`?

Same goes for a couple places below where flows-app-production is hardcoded.

src/store/corestore.ts
```diff
@@ -21,16 +21,12 @@ import {
   User,
   UserRole
 } from "../sharedtypes";
+import firebaseConfig from "./firebaseconfig.json";
```
balderdash: Would this file live in lastpass?

goldjay: Yes, that's the plan.

functions/src/index.ts
goldjay: Turns out no. Looks like the gcloud project id is available in functions so I'll just use that.

PR #168, opened 2020-08-13
Collect firestore accessors into a single object

balderdash:
We had a file called `corestore.ts` that had a bunch of functions for talking to firestore. This PR moves those functions into a single datastore accessor object as a first step to being able to swap out the data store with Maisha's REST API. Which backend to use is configurable in `store/config.ts`, and that config is passed into `initializeStore()` in `transport/datastore` which creates the appropriate datastore object.

I'll be further trimming down the number of methods available on `DataStore` in upcoming PRs, I wanted to keep this PR as straightforward as possible a refactor of `corestore.ts` as I could.


package.json
```diff
@@ -35,6 +35,7 @@
     "start:react": "react-scripts start",
     "build": "react-scripts build; git rev-list -n 1 HEAD > build/about.txt; date >> build/about.txt",
     "test": "react-scripts test",
+    "typecheck": "tsc --noEmit",
```
yrichard, line 38: why is that here?

src/App.tsx
```diff
@@ -1,16 +1,19 @@
-import firebase from "firebase/app";
 import "firebase/auth";
```
yrichard, line 1: not for this diff, but if this is a public repo, we should add a copyright header to all files.

package.json
balderdash: I like to use `yarn typecheck` to sanity check refactoring changes. It's quicker than a full build.

PR #174, opened 2020-10-15
AUD-1738 Use Maisha API

balderdash:
Fixes: https://auderenow.atlassian.net/browse/AUD-1738

This PR sets up the connection to the Maisha API server.

`transport/maisha.ts` defines simple wrappers around their endpoints with properly typed parameters and responses. It also handles saving and re-sending the authorization token you get when you sign in.

`transport/maishDatastore.ts` uses the class above to implement a `DataStore`, which provides all the functions expected by the rest of the UI code. It handles the translation between the `Task` and `Claim` objects used throughout the codebase and the `care_pathway_instance` objects we get back from maisha.

**Reviewer should merge**
* No


src/Components/ListItem.tsx
```diff
@@ -33,7 +32,7 @@ export class ListItem extends React.Component<{
           </span>
         </div>
         <div>
-          {"Total Reimbursement: " + dataStore.formatCurrency(claimsTotal)}
+          {"Total Reimbursement: " + formatCurrency(claimsTotal)}
```
yrichard, line 35: I don't see where this function is imported from.

src/transport/maishaDatastore.ts
```diff
@@ -0,0 +1,373 @@
...
+          this.authStateChangedCallbacks.map(cb => cb(false));
+        }
+        try {
+        } catch {}
+      })();
+    }
+  }
+
+  async login({ username, password }: { username: string; password: string }) {
+    const res = await fetch(this.endpointRoot + "/users/sign_in", {
```
yrichard, line 63: We don't support IE, right? fetch isn't implemented there.

src/transport/maishaDatastore.ts
```diff
@@ -0,0 +1,373 @@
...
+    newState: TaskState,
+    notes: string,
+    payment?: PaymentRecord
+  ): Promise<void> {
+    if (tasks.length === 0) {
+      return;
+    }
+    const entries = tasks.map(task => task.entries).flat();
+    const reviewedEntries = reviewedTasks.map(task => task.entries).flat();
+    const flaggedEntries = flaggedTasks.map(task => task.entries).flat();
```
yrichard, line 109: not a biggie but I would have just made copies of entries, either using slice() either using [...entries]

src/transport/maishaDatastore.ts
```diff
@@ -0,0 +1,373 @@
...
+    reviewedTasks: Task[],
+    flaggedTasks: Task[],
+    newState: TaskState,
+    notes: string,
+    payment?: PaymentRecord
+  ): Promise<void> {
+    if (tasks.length === 0) {
+      return;
+    }
+    const entries = tasks.map(task => task.entries).flat();
```
yrichard, line 107: why not adding .map(entry => entry.claimID) since that's what you need?

src/transport/maishaDatastore.ts
```diff
@@ -0,0 +1,373 @@
...
+      paidStatus,
+    };
+  }
+
+  async loadTasks(taskState?: TaskState): Promise<Task[]> {
+    const claimFilters = this.getClaimQuery(taskState);
+    const { facilities } = await this.maishaApi.facilities();
+    return (
+      await Promise.all(
+        facilities.map(async facility => {
```
yrichard, line 278: This is going to be potentially expensive, how many of those can potentially run in parallel?

src/transport/maisha.ts
```diff
@@ -0,0 +1,253 @@
...
+      "POST",
+      body
+    );
+  }
+
+  async facilities(): Promise<{ facilities: MaishaFacility[] }> {
+    return await this.fetchWithToken("/facilities?loyalty_enabled=true");
+  }
+
+  async carePathwayInstances(
```
yrichard, line 176: maybe "fetchCarePathwayInstances" or "searchCarePathwayInstances" ? I feel it's better naming to be explicit about what action a function does. So maybe the same for "facilities", "loyaltyPayment" etc...

src/transport/maishaDatastore.ts
yrichard: I'm not going to block on this, because it would probably require a new API from Maisha, but maybe add a TODO or a note about perf.

src/transport/maishaDatastore.ts
```diff
@@ -0,0 +1,373 @@
...
+  ClaimEntry,
+  PaymentRecord,
+  Site,
+  Task,
+  TaskState,
+  UserRole,
+} from "../sharedtypes";
+import { DataStore, Flag } from "./baseDatastore";
+
+import moment from "moment";
```
yrichard, line 19: I didn't see the moment api in the package.json. Are you sure it's not globally added to your environment?

yrichard (approved):
Looks good overall, just one missing import and a few nits that could probably wait.

src/transport/baseDatastore.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  abstract loadFlags(tasks: Task[]): Promise<{ [taskId: string]: Flag[] }>;
+
+  abstract setClaimNotes(
+    task: Task,
+    claimIndex: number,
+    notes: string
+  ): Promise<void>;
+
+  // Optional Methods with no-op default implementations
```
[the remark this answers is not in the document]
balderdash: They were for features Maisha didn't need, some of the corresponding UI elements just don't show up if these methods don't return anything and others are getting hidden explicitly.

src/transport/maishaDatastore.ts
balderdash: That's right.

src/transport/maishaDatastore.ts
balderdash: This is actually what I'm working on now, they delivered the necessary API last night.

src/transport/maishaDatastore.ts
balderdash: Looks like it was indirectly required by another package, I added the direct dependency.

PR #176, opened 2020-10-15
AUD-1829 Bump react-scripts version, fix some scrolling and react-router issues

balderdash:
Fixes https://auderenow.atlassian.net/browse/AUD-1829

Upgraded react-scripts to get optional chaining support in typescript, some CSS fixes to avoid having 3 nested divs each with their own vertical scrollbars, and fixing a bug with routing to the admin tab.

Most of then lines changed are just `yarn.lock` changes from the react-scripts upgrade.

**Reviewer should merge**

* No

### DIFF
--- a/src/Screens/AuditorPanel.tsx
+++ b/src/Screens/AuditorPanel.tsx
@@ -234,7 +234,7 @@ export class AuditorDetails extends React.Component<
     }
 
     return (
-      <LabelWrapper key={JSON.stringify(entry + "_" + index)}>
+      <LabelWrapper key={JSON.stringify("entry_" + index)} disableScroll={true}>
         <div className="mainview_padded">
           <div className="mainview_row">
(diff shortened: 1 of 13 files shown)


src/Screens/AuditorPanel.tsx
```diff
@@ -234,7 +234,7 @@ export class AuditorDetails extends React.Component<
     }
 
     return (
-      <LabelWrapper key={JSON.stringify(entry + "_" + index)}>
+      <LabelWrapper key={JSON.stringify("entry_" + index)} disableScroll={true}>
```
yrichard, line 237: using index isn't the best, as it might lead to some error. I'm not familiar with this UI, so it might not be a problem but if you have access to the patient ID, use that instead.

yrichard (approved):

PR #187, opened 2020-10-16
AUD-1736 Hide some buttons that don't work

balderdash:

The Maisha API doesn't support everything that was in the firebase version, so these buttons don't work. Hide them.

**Reviewer should merge**
* No


jcmorrow (approved):
Looks nice and hidden ⬛

PR #191, opened 2020-10-16
AUD-1736 Fix search behavior

balderdash:

Make search actually case-insensitive (it previously never matched a search term if it contained capital letters), and split terms using spaces instead of commas.

**Reviewer should merge**
* No

### DIFF
--- a/src/util/search.ts
+++ b/src/util/search.ts
@@ -7,7 +7,13 @@ export interface DateRange {
   endDate: Moment | null;
 }
 
-const DEFAULT_FILTERS_ENTRY: (keyof ClaimEntry)[] = [
+const CLAIM_SEARCH_KEYS_MAP: { [key: string]: (keyof ClaimEntry)[] } = {
+  patient: ["patientFirstName", "patientLastName"],
+  item: ["items"],
+  phone: ["phone"],
+};
+
+const ALL_CLAIM_SEARCH_KEYS: (keyof ClaimEntry)[] = [
   "patientFirstName",
   "patientLastName",
   "items",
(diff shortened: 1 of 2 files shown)


src/util/search.ts
```diff
@@ -32,25 +34,18 @@ export const containsSearchTerm = (
...
       }
     } else {
-      const defaultFilters = entry.hasOwnProperty("claimedCost")
-        ? DEFAULT_FILTERS_ENTRY
-        : DEFAULT_FILTERS_HISTORY;
       if (
-        defaultFilters.some(filter => {
+        DEFAULT_FILTERS_ENTRY.some(filter => {
+          if (entry[filter]) {
+          }
```
jcmorrow, line 45: What's going on here?

src/util/search.ts
```diff
@@ -32,25 +34,18 @@ export const containsSearchTerm = (
       const value = keyValue[1].toLowerCase().trim();
       if (
         entry[key] &&
-        entry[key]
-          .toString()
-          .toLowerCase()
-          .includes(value)
+        entry[key].toString().toLowerCase().includes(value.toLowerCase())
```
jcmorrow, line 37: Do we say on the UI that searching by a specific field is an option? This seems really useful.

src/util/search.ts
```diff
@@ -5,8 +5,16 @@ export interface DateRange {
   endDate: Moment | null;
 }
 
-const DEFAULT_FILTERS_ENTRY = ["patient", "item", "pharmacy"];
-const DEFAULT_FILTERS_HISTORY = ["id", "time", "description", "notes"];
+const DEFAULT_FILTERS_ENTRY = [
+  "patientFirstName",
+  "patientLastName",
+  "item",
```
jcmorrow, line 11: Is "item" like sold product?

jcmorrow (approved):
A couple of small questions, but overall this looks good 🔍

src/util/search.ts
balderdash: That's right.

src/util/search.ts
balderdash: It shows up in the tooltip when you expand the search option in the pharmacy list, but yeah it's not super discoverable.

src/util/search.ts
balderdash: This is the "search without specifying a key" logic, i.e. look through all the available keys for a match. I just cleaned this up a bit.

PR #199, opened 2020-10-17
Render claim notes

balderdash:

Notes is now a list, not a single string, so render it appropriately. Notes can be multiline, so I added a border between notes.

**Screenshots**
image

**Reviewer should merge**
* No

### DIFF
--- a/src/Components/ClaimNotes.tsx
+++ b/src/Components/ClaimNotes.tsx
@@ -9,12 +9,12 @@ import { dataStore } from "../transport/datastore";
 interface Props {
   claimIndex: number;
   task: Task;
-  notes: string;
+  notes: string[];
   cannedNotes: string[];
 }
 
 interface State {
-  notes: string;
+  newNote: string;
   editing: boolean;
   cannedClaimNotes?: string[];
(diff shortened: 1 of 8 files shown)


src/Components/ClaimNotes.tsx
```diff
@@ -49,24 +49,40 @@ class ClaimNotes extends React.Component<Props, State> {
...
-      <div className="claimnotes_row">
+      <div>
+        {this.props.notes.length > 0 && (
+          <>
+            <div>Notes:</div>
+            <div>
+              {this.props.notes.map((note: string) => (
+                <div className="claimnotes_note">
+                  {note.split("\n").map(line => (
+                    <div>{line}</div>
```
jcmorrow, line 73: Maybe it would be more semantic to use `<p>` tags here?

jcmorrow (approved):
One comment, otherwise - looks good 🥃

src/Components/ClaimNotes.tsx
balderdash: Makes sense!

PR #202, opened 2020-10-19
Collapse patients after selecting an action

balderdash:

**Screenshots**
collapse

**Reviewer should merge**
* No

### DIFF
--- a/src/Components/LabelWrapper.tsx
+++ b/src/Components/LabelWrapper.tsx
@@ -9,9 +9,17 @@ interface Props {
   renderLabelItems?: () => JSX.Element;
   searchPanel?: JSX.Element;
   disableScroll?: boolean;
+  collapsible?: boolean;
+  collapsed?: boolean;
+  onCollapse?: (collapsed: boolean) => void;
+  collapsedDisplay?: React.ReactNode;
 }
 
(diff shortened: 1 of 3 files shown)


src/Components/LabelWrapper.tsx
```diff
@@ -9,9 +9,17 @@ interface Props {
...
   disableScroll?: boolean;
+  collapsible?: boolean;
+  collapsed?: boolean;
+  onCollapse?: (collapsed: boolean) => void;
+  collapsedDisplay?: React.ReactNode;
 }
 
 class LabelWrapper extends Component<Props> {
+  _onCollapse = () => {
+    this.props.onCollapse && this.props.onCollapse(!this.props.collapsed);
```
jcmorrow, line 20: It's a little strange this this param is called `collapsed` but the value that we are passing in is `!collapsed`. Maybe one of these should be renamed?

jcmorrow (approved):
One comment on the naming of the param, but otherwise this looks awesome 😎

src/Components/LabelWrapper.tsx
balderdash: This is the method that toggles the collapsed state, hence the negation. I kept the parameter name but renamed the method to make that a bit clearer.

PR #204, opened 2020-10-20
Add a history link

balderdash:

Add a history link for claims.

**Screenshots**

image

### DIFF
--- a/src/Components/ClaimNotes.tsx
+++ b/src/Components/ClaimNotes.tsx
@@ -1,6 +1,6 @@
 import "./ClaimNotes.css";
 
-import React, { ChangeEvent, Fragment } from "react";
+import React, { ChangeEvent } from "react";
 
 import Button from "./Button";
 import { Task } from "../sharedtypes";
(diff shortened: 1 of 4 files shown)


src/Components/ClaimNotes.tsx
```diff
@@ -59,42 +59,51 @@ class ClaimNotes extends React.Component<Props, State> {
...
-            <Button
-              className="claimnotes_button"
-              label="Edit"
-              onClick={this._onEdit}
-            />
-          </Fragment>
+      <div className="claimnotes_wrapper">
+        <div className="claimnotes_row">
+          {editing ? (
+            <>
```
jcmorrow, line 68: What is this? I've never seen `<>` in react before.

jcmorrow (approved):
One little question on the `<>` element, but otherwise 👍

src/Components/ClaimNotes.tsx
liangmicha: I think it's the new shorthand for <Fragment>: https://reactjs.org/docs/fragments.html#short-syntax

PR #209, opened 2020-10-30
Show note in the current note section if the POST succeeded

jcmorrow:
Right now you won't see the note until you either refresh all claims manually by refreshing the page, or otherwise trigger a refresh of all claims in the app (both of which are pretty painful and slow). Let's add the note to the currently displayed ones if the POST request was a successful response, otherwise, pop an error so the user doesn't expect that note to be saved.


src/Components/ClaimNotes.tsx
```diff
@@ -40,8 +42,17 @@ class ClaimNotes extends React.Component<Props, State> {
...
+      () => {
+        this.setState({
+          editing: false,
+          notes: [...this.state.notes, newNote],
+        });
+      },
+      () => {
+        alert("There was an error saving your note");
+      }
+    );
```
balderdash, line 55: Nit: With `await`, this could just be:
```
try {
  await dataStore.setClaimNotes(...);
} catch {
  alert(...);
  return;
}
this.setState({...});
```
The behavior should be identical though, it's just syntactic sugar.

src/transport/maisha.ts
```diff
@@ -237,7 +238,7 @@ export class MaishaApi {
       message: string;
     };
   }): Promise<void> {
-    await this.fetchWithToken("/review_note", "POST", body);
+    return await this.fetchWithToken("/review_note", "POST", body);
```
balderdash, line 241: With just `await`, `postReviewNote` already returns a promise that only resolves once the `fetchWithToken()` promise has resolved. The `return` is unnecessary unless you're trying to return the value (i.e. the API response) as well.

balderdash (approved):
Just a couple minor comments about `await`, but the change itself looks good!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FJQQwgVUMDIyAM%2Fgiphy.gif&token=great" data-gitmeme-token="great"></a>

src/transport/maisha.ts
jcmorrow: Ah, gotchya, thanks!

src/Components/ClaimNotes.tsx
jcmorrow: Hmmmmm yeah, I think the thing I don't love about that is that `catch` could be catching all sorts of things, not just a problem with the network request. It feels a little like verging on using exceptions for control flow (but that's kind of what Promises do to begin with, it's hard to avoid).

src/transport/maisha.ts
```diff
@@ -193,7 +194,7 @@ export class MaishaApi {
   }
 
   async getFacilities(): Promise<{ facilities: MaishaFacility[] }> {
-    return await this.fetchWithToken("/facilities?loyalty_enabled=true");
+    await this.fetchWithToken("/facilities?loyalty_enabled=true");
```
balderdash, line 197: For the `get...` endpoints you'll want to leave the `return` in place because we actually do care about the API responses.

PR #211, opened 2020-11-18
Add transaction id to individual claim view

liangmicha:
Problem:
- Instead of ops team telling us Transaction "Amani Medical Clinic/11-14 11:33/15-Nov/Oraquick/", they can tell us transaction ID when trying to refer to something on the Audere dashboard


liangmicha:
oops sorry, wrong repo

